PR #5, opened 2020-01-09
Organization and reorganization

aimalz:
This PR is for the basic setup of the repo and organization of code.


sschmidt23 (approved):
Changes look good, degradation being merged back in to creation makes a bit more sense than how it was before.

PR #6, opened 2020-01-10
Estimation module template

aimalz:
This branch will be ready to merge after some clean-up.


aimalz:
We're going to remove the ceci dependency to have a simpler estimator superclass whose subclasses can wrap photo-z PDF codes.

aimalz:
Last things to do before merge: 
- [ ] Combine `tomo.py` and `base.py` (and credit [tomo_challenge](https://github.com/LSSTDESC/tomo_challenge))
- [ ] Move paths and output format parameters from `base.py` to` base.yaml`; make estimators read in both their own param file and `base.yaml` (with path names that we will change for different test conditions)
- [ ] Fill in documentation
- [ ] Remove obsolete files, simplify directory structure, change various names

johannct:
random.py should be renamed as random_pz.py for instance, as I am not convinced the namespace can be safely guaranteed.

johannct:
`from estimator import Estimator as BaseEstimation` : I think it should be `from ..estimator etc` to be ok with local imports

johannct:
in estimator.py the files are loaded at initiallization, instead of when that are actually used..... I think we should be lazier.... Not sure how this works out with the parallel_io branch effort

johannct:
I am opening the pull request for review as it seems that github can help automatically detect relevant templates for actions, but this is based on what is on master. CI is the last thing to deal with in this first stage overall construction of the package

rail/estimation/README.md
```diff
@@ -2,13 +2,30 @@
...
 We will have to make a decision about the acceptable output format(s) of redshift posteriors.
 
+## Structure
+
+Examples of wrapped codes can be found as submodules of `rail.estimation.algos`.
+Each must correspond to a config file in with any parameters the method needs, temporarily located in `rail.estimation.configs` but not actually a code directory.
+. . .
+
+## Usage
+
```
jbkalmbach: Specify that this is in the examples directory?

rail/estimation/README.md
```diff
@@ -2,13 +2,30 @@
...
 
 For the sake of this challenge, we will run scripts that accept test set photometry, run a particular pre-trained photo-z estimation code, and produce estimated photo-z posteriors.
 Where possible, we wil use formats compatible with other LSST-DESC pipelines, including [TXPipe](https://github.com/LSSTDESC/TXPipe/).
 Code here will provide a script template for wrapping a machine learning code that we will run automatically on a variety of test sets blinded from those who submit scripts.
 We will have to make a decision about the acceptable output format(s) of redshift posteriors.
 
+## Structure
+
+Examples of wrapped codes can be found as submodules of `rail.estimation.algos`.
+Each must correspond to a config file in with any parameters the method needs, temporarily located in `rail.estimation.configs` but not actually a code directory.
```
jbkalmbach: Point to example config file?

examples/main.py
```diff
@@ -0,0 +1,54 @@
...
+        base_config = 'base.yaml'
+        input_yaml = argv[1]
+    elif len(argv) == 3:
+        base_config = argv[1]
+        input_yaml = argv[2]
+    else:
+        print(len(argv))
+        print("Usage: main <yaml file>")
+        sys.exit()
+    name = input_yaml.split("/")[-1].split(".")[0]
```
jbkalmbach, line 21: Should the name come from this? Doesn't this mean if you wanted to run the same algo with different configs it won't work? Perhaps the name should be inside the config file?

Actually it looks like there are lines in the config files for simpleNN and FZBoost (but not random) for class name and output file already so I think the changes need to be made in estimator.py and maybe not here. I'll leave a note there too.

rail/estimation/estimator.py
```diff
@@ -0,0 +1,86 @@
...
+                                                self.groupname)
+        self.testfile = base_dict['testfile']
+        self.num_rows = get_input_data_size_hdf5(self.testfile, self.groupname)
+        self._chunk_size = base_dict['chunk_size']
+
+        self.test_fmt = self.testfile.split(".")[-1]
+        # self.test_data = load_data(self.testfile, self.test_fmt)
+        # move reading of test data to main.py so we can loop more easily
+
+        self.code_name = type(self).__name__
```
jbkalmbach, line 55: This should use settings in `run_params` of config file.

rail/estimation/estimator.py
```diff
@@ -0,0 +1,86 @@
...
+        self.saveloc = os.path.join(self.outpath, self.code_name + '.hdf5')
```
jbkalmbach, line 56: This should also use settings in `run_params`.

rail/estimation/estimator.py
```diff
@@ -0,0 +1,86 @@
...
+
+        self.config_dict = config_dict
+
+    def inform(self):
+        """
+        A training algorithm for the individual photo-z method, should be
+        implemented in the subclass
+        """
+        pass
```
jbkalmbach: raise NotImplementedError here in the base class to force this to be defined in subclass?

rail/estimation/estimator.py
```diff
@@ -0,0 +1,86 @@
...
+        Returns:
+        --------
+        pz_dict:
+          dictionary of output photo-z params, must include zmode and pdf
+          note: zgrid will still be a class variable for now
+
+        should create photo-z estimates with set names, TBD
+        for demo will just be `z_mode`
+        """
+        pass
```
jbkalmbach, line 86: raise NotImplementedError here in the base class to force this to be defined in subclass?

setup.py
```diff
@@ -0,0 +1,34 @@
+# from distutils.core import setup
```
jbkalmbach: Clean up commented lines?

jbkalmbach (approved):
Tested all three example codes and it looks good. Tests run and it looks like it works with Github actions which is really cool to see.

Didn't really leave comments on things that are in the list of things still to work on in the immediate future.

Biggest comment is on using the `run_params` information better so that it is easier to run multiple iterations of the same code with different settings if we want.

rail/estimation/README.md
johannct: 1be290c

rail/estimation/estimator.py
johannct: 5788cc1

rail/estimation/README.md
johannct: 5788cc1

rail/estimation/estimator.py
johannct: indeed

johannct:
I just added a basic coverage running, and it seems to work

examples/main.py
johannct: I would like to move resolution of this outside of this pull request, as the code is not there at all to honor the config class_name attribute. @jbkalmbach do you agree?

johannct:
remaining issue with class_name moved to new issue #22

PR #9, opened 2020-01-16
creation baseline

jbkalmbach:
Open up this PR to track the work on the creation end. This endpoint for this PR I think is when we have a fully functional RAIL creation notebook.


rail/creation/creator.py
```diff
@@ -0,0 +1,63 @@
...
+                          'gamma_z':0.039,
+                          'gamma_y':0.039,
+                          'm5_u':28.1,
+                          'm5_g':29.4,
+                          'm5_r':29.5,
+                          'm5_i':28.8,
+                          'm5_z':28.1,
+                          'm5_y':26.9}
+
+            for band in ['u','g','r','i','z','y']:
```
jbkalmbach: Do we want to hard code the bands and the`err_params`? Do we want to just move forward with this and leave an issue to come back to this?

rail/creation/realNVP.py
```diff
@@ -0,0 +1,297 @@
...
+                  fraction_masked))
+        self._num_masked = None
+        self._fraction_masked = float(fraction_masked)
+        self._reverse_mask = self._fraction_masked < 0
+
+      if shift_and_log_scale_fn is not None and bijector_fn is not None:
+        raise ValueError('Exactly one of `shift_and_log_scale_fn` and '
+                         '`bijector_fn` should be specified.')
+
+      if shift_and_log_scale_fn:
```
jbkalmbach: Do we need this method that uses`affine_scalar`? If so, we need to add the import for `affine_scalar`. If not, let's remove this.

rail/creation/creator.py
jfcrenshaw: That stuff was when I was just initially playing around with the error convolution. I think we should come back to it later, and either leave it or delete it for now.

jbkalmbach: Actually I was looking at it and we could put it in this params dictionary that you specify in the `__init__`

rail/creation/realNVP.py
jfcrenshaw: I don't think we need `affine_scalar`. That `ValueError` says that we can't use `shift_and_log_scale_fn` if we use `bijector_fn`. Since we're building normalizing flows from quadratic splines (or potentially some other bijector we supply), we will always be specifying `bijector_fn`.

jbkalmbach: Ok, removing this.

rail/creation/normalizingFlow.py
```diff
@@ -0,0 +1,259 @@
...
+            params = save_dict['params']
+            transform_data = save_dict['transform_data']
+            inv_transform_data = save_dict['inv_transform_data']
+        
+        # make sure the hyperparameter dict contains nfeatures
+        if 'nfeatures' not in hyperparams:
+            raise KeyError('nfeatures must be in the hyperparameter dictionary')
+        
+        # save the hyperparameters
+        self.hyperparams = hyperparams.copy()
```
johannct: There is  no need to copy over here

rail/creation/normalizingFlow.py
```diff
@@ -0,0 +1,259 @@
...
+        
+        # create flax modules
+        @nn.module
```
johannct: there is no check that `module` is usable here, we probably need one.

rail/creation/normalizingFlow.py
```diff
@@ -0,0 +1,259 @@
...
+        def inverse_module(x):
+            return module(hyperparams).bijector.inverse(x)
+        @nn.module
+        def sampler_module(n_samples, seed):
+            return module(hyperparams).sample(n_samples, seed)
+        @nn.module
+        def log_prob_module(x):
+            return module(hyperparams).log_prob(x)
+
```
johannct: I would put the following before the aux function definitions, as it concerns dealing with the __init__ inputs

rail/creation/normalizingFlow.py
```diff
@@ -0,0 +1,259 @@
...
+        # save the functions for data transformation
+        self.transform_data = transform_data
+        self.inv_transform_data = inv_transform_data
+        
+        
+    def forward(self, y):
+        x = self._forward(y)
+        if self.inv_transform_data:
+            x = self.inv_transform_data(x)
+        return x
```
johannct: The condition inside the call could be a bit costly if the function is called very often. It may be better considering a trivial identity lambda function if arg inv_transform_data is None (note that semantically the code is stretching a bit python laziness : if predicate should be True or False, not None).

rail/creation/normalizingFlow.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def inverse(self, x):
+        trans_x = x
+        if self.transform_data:
```
johannct: likewise, a lambda identity function may be preferable here and elsewhere

rail/creation/normalizingFlow.py
jfcrenshaw: So do you mean if no transforms are passed on initialization, set the transform functions to be `lambda x: x`, then always call the transform functions when doing log_prob, etc?

@johannct : exactly, and you can do that once and for all in the __init__, so the conditions disappear from the function calls (generically preferable I believe)

sjoudaki:
A few notes on trying to run the notebooks in the branch issue/7/creation-baseline (with help from @jfcrenshaw and @aimalz):

Trying to run the notebooks in rail/creation directory (specifically, twomoons_flow_example.ipynb, creation_with_selection_function.ipynb, creation_demo.ipynb), we arrive at the error

AttributeError: module 'tensorflow' has no attribute 'distributions'

when trying to execute

"from tensorflow_probability.python.internal.backend import jax as tf"

This occurs in a setup where: 
i) jupyter.nersc is used
ii) the rail-creation environment is activated (via "conda env create -f environment.yml", then "conda activate rail-creation", then "python -m ipykernel install --user --name rail-creation --display-name railcrea")
iii) rail is installed (via "python setup.py develop --user")
iv) yaml is installed (via "conda install yaml")
v) "realNVP" instead of "jax_nf.real_nvp" (in twomoons_flow_example.ipynb)

John Franklin notes that he does not obtain the error when not installing rail. This needs to be understood better.

On a separate note, when trying to run redshift_flow_example.ipynb when using the railcrea kernel, we obtain

ModuleNotFoundError: No module named 'GCRCatalogs'

However, this error does not appear when using the desc-python kernel instead (while the desc-python kernel is not adequate for the other parts of the notebook).

jfcrenshaw:
> A few notes on trying to run the notebooks in the branch issue/7/creation-baseline (with help from @jfcrenshaw and @aimalz):
> ....

Just some clarifying comments on the above:
- the creation branch works fine when you just install all of the modules, but WITHOUT running `python setup.py develop --user`. Running that command seems to cause problems with tensorflow, even when you add the correct version numbers to setup.py. I am not sure why this happens.
- The GCR catalog is in the DESC stack, so to run the catalog loader in `redshift_flow_example.ipynb`, you have to be in an environment that can see the DESC stack. The best way to do this is to [make your own clone of the desc environment](https://github.com/LSSTDESC/desc-python/wiki/Add-Your-Own-Packages-to-the-desc-python-Environment), then install your packages in that environment. (If you don't want to do that, you can hack around it by running the GCR catalog loader in the shared desc environment, saving the loaded data (e.g. pickling it), and then loading the data from that saved file when you run the notebook in an environment without access to the GCR catalog.)

johannct:
> 
>     * the creation branch works fine when you just install all of the modules, but WITHOUT running `python setup.py develop --user`. Running that command seems to cause problems with tensorflow, even when you add the correct version numbers to setup.py. I am not sure why this happens.
> 

It is critical to discourage use of `--user` anyway. Documentation should not mention it, and if dedicated path is needed `--prefix` is to be preferred. Given the fact that we eventually will likely move to conda/conda-forge as our primary system of package management and installation, conda environment are anyway the baseline path to follow, and there is no reason to use `--user` in that framework.

>     * The GCR catalog is in the DESC stack, so to run the catalog loader in `redshift_flow_example.ipynb`, you have to be in an environment that can see the DESC stack. The best way to do this is to [make your own clone of the desc environment](https://github.com/LSSTDESC/desc-python/wiki/Add-Your-Own-Packages-to-the-desc-python-Environment), then install your packages in that environment. (If you don't want to do that, you can hack around it by running the GCR catalog loader in the shared desc environment, saving the loaded data (e.g. pickling it), and then loading the data from that saved file when you run the notebook in an environment without access to the GCR catalog.)

Note that GCR/gcr-catalogs are both pip installable, and recent jupyter version ship with a handy `%pip` magic.

jfcrenshaw:
I moved all of the normalizing flow stuff to the [pzflow package](https://github.com/jfcrenshaw/pzflow), where I personally implemented all of the flow and neural spline couplings myself. This has numerous benefits, including faster flows, more transparent code, and removing the dependencies on tensorflow, tensorflow probability, and flax.

jfcrenshaw:
By next week, I am planning to fix the pytest failure, add pytest coverage of the GCR Catalog Loader, and add example notebooks with degraders and the Catalog Loader. I think at that point, this branch will be ready to be merged into master.

jfcrenshaw:
The only thing left is to write tests for the GCR Catalog Loader

jfcrenshaw:
I believe this pull request is essentially complete with respect to the creation module (aside from editing the README, which I will do soon) -- it would be great if someone else reviewed the code and let us know if they disagree with any of the choices Bryce and I made.

I am also going to add one more example notebook that integrates the creation module with the estimation module.

rail/creation/creator.py
```diff
@@ -0,0 +1,102 @@
...
+                )
+                new_sample = self.degrader(new_sample, seed=rng.integers(1e18))
+                # add these to the larger set
+                outputs = pd.concat((outputs, new_sample), ignore_index=True)
+            # cut out the extras
+            outputs = outputs[:n_samples]
+
+        # calculate posteriors
+        if include_pdf:
+            posteriors = self.generator.pz_estimate(outputs, grid=pz_grid)
```
aimalz: Could this be changed to access the `rail.creation.Generator.pz_estimate()` method from within `rail.creation.Creator`? This has been requested offline by @johannct and @sschmidt23 to enable the production of posteriors/likelihoods at a static set of photometric/redshift data points, respectively, rather than those sampled from the model of the joint distribution. I think this use case would also apply to [the degradation demo](https://github.com/LSSTDESC/RAIL/blob/issue/7/creation-baseline/examples/creation/degradation-demo.ipynb) if it went a few steps further (not that the demo has to!), where a naive user whose `rail.estimation.Estimator` is a `rail.creation.Creator` would create their own estimated model p(z, photometry) from the training set drawn from the degraded model and then evaluate posteriors within that model at the test set photometry drawn from the original model they didn't know about.

rail/creation/flow_wrapper.py
```diff
@@ -0,0 +1,12 @@
+from rail.creation.generator import Generator
+
+
+class FlowGenerator(Generator):
```
aimalz: Would you mind adding a comment to explain why there's another layer of abstraction for `Generator` vs. `FlowGenerator`? Also, is `FlowGenerator` only used for testing? I can't quite follow how it relates to the `Creator`.

rail/creation/generator.py
```diff
@@ -0,0 +1,17 @@
+# Abstract base class defining a generator,
+# which represents a probability distribution
+
+from abc import ABC, abstractmethod
+
+
+class Generator(ABC):
```
aimalz: I have a perhaps irrelevant concern about naming confusion with "Generator" meaning something specific in Python. Could we call this "Creator_gen" or something like what was done for [qp's `qp.Pdf_gen` class](https://github.com/LSSTDESC/qp/blob/master/qp/pdf_gen.py)? (This question might be degenerate with my confusion about `FlowGenerator` vs. `Generator`, sorry.)

rail/creation/generator.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self):
+        """Initialize Generator that can sample galaxy data and draw redshift posteriors"""
+
+    @abstractmethod
+    def sample(self, n_samples, seed=None, **kwargs):
+        """Return a random sample of the distribution with size n_samples"""
+
+    @abstractmethod
+    def pz_estimate(self, data, grid, **kwargs):
```
aimalz: Another potential instance of naming confusion might be the `rail.creation.Generator.pz_estimate()` method, since we use "estimate" to mean something else in the `rail.estimation` subpackage. Would `get_posterior()` be an acceptable substitute? (A more generic version of this might be `get_marginal()` where a user would specify all dimensions except one, with a grid for the omitted one, and then `get_posterior()` would be understood to be like a special instance of `get_marginal(data=_ugrizy_, grid=_redshifts_)`, or something like that.)

rail/creation/creator.py
jfcrenshaw: I created a standalone `get_posterior(data, column, grid)` method that calculates posteriors for the given column over the values in the given grid. The `sample` method still has an `include_pdf` flag, that now calls `get_posterior` with `column='redshift'`.

rail/creation/flow_wrapper.py
jfcrenshaw: `Generator` is an abstract class that shows the required methods for an object to be the generator in a `Creator` object. `FlowGenerator` is a subclass of `Generator` that wraps a pzflow `Flow` object to be a `Generator` object. If you want, you can think of an analogy with the Estimation module. `Generator` is like the base estimator, and `FlowGenerator` is like one of the pz estimation codes that wrap the base estimator (e.g. flexzboost). I have added some comments to these files to indicate this relationship. However, I have renamed Generator -> Engine (see comment below).

rail/creation/generator.py
jfcrenshaw: In keeping with the train analogy (which I think is the most important design principle of RAIL?), I have renamed `Generator` to `Engine`.

rail/creation/generator.py
jfcrenshaw: I have renamed this method to `get_posterior(data, column, grid)`. Works the same as before, except now with the `column` parameter that indicates which column to calculate posteriors for. This method is also directly accessible from the `Creator` object (see comment above).

rail/creation/creator.py
```diff
@@ -0,0 +1,117 @@
...
+                # add these to the larger set
+                outputs = pd.concat((outputs, new_sample), ignore_index=True)
+            # cut out the extras
+            outputs = outputs[:n_samples]
+
+        # calculate posteriors
+        if include_pdf:
+            posteriors = self.get_posterior(outputs, column="redshift", grid=pz_grid)
+            outputs.attrs["pz_grid"] = pz_grid
+            outputs["pz_pdf"] = list(posteriors)
```
aimalz, line 115: I just noticed that the outputs here don't match [those of the estimators](https://github.com/LSSTDESC/RAIL/blob/issue/7/creation-baseline/rail/estimation/estimator.py), which could cause problems for evaluators seeking to make a direct comparison. I propose the following that would be shared among both: 
- `z_pt` for a Creator's true redshift (which I think is currently taken directly from the input data, i.e. `redshift_true` for the Buzzard demo data) and an Estimator's "best" point estimate (which is currently `zmode` but should be left generic for codes that have a different summarization, since we can always impose our own from the output posteriors)
- `z_posterior_fmt` for the parameterization of the output posteriors, in this case ["interp_pdf"](https://github.com/LSSTDESC/qp/blob/master/qp/interp_pdf.py) for all current Creator and Estimator subclasses
- `z_posterior_glob` for a Creator's current `pz_grid` and an Estimator's current `zgrid`
- `z_posterior_cat` for what's currently `pz_pdf` for both the Creator and Estimator superclasses

While those would ultimately be replaced with `z_posterior` for the qp.Ensemble defined by its parameterization, catalog-level metaparameters, and per-galaxy parameters (and possibly `z_pt` and the photometry columns as metadata files?) after #33 is resolved), I don't see the harm in editing both rail/creation/creator.py and rail/estimation/estimator.py (plus the tests for both) in this branch to make these match. (If this seems too far out of scope, however, feel free to punt this entire comment to a new issue, just so we don't forget to fix it before the evaluation dev branch advances too much further.)

rail/creation/flow_wrapper.py
sschmidt23: To reduce some confusion on this point, I think we should do what is listed as a comment on the top of flowEngine.py and move this to a subdirectory (engines/ ?).  I think that might make it more clear that flowEngine is one specific engine that wraps pzflow.

rail/creation/creator.py
sschmidt23: I think the scope of this PR is already large enough, I suggest that we create a separate issue/PR on output formats and deal with this after we merge this base branch.

rail/creation/creator.py
```diff
@@ -0,0 +1,117 @@
...
+                new_sample = self.degrader(new_sample, seed=rng.integers(1e18))
+                # add these to the larger set
+                outputs = pd.concat((outputs, new_sample), ignore_index=True)
+            # cut out the extras
+            outputs = outputs[:n_samples]
+
+        # calculate posteriors
+        if include_pdf:
+            posteriors = self.get_posterior(outputs, column="redshift", grid=pz_grid)
+            outputs.attrs["pz_grid"] = pz_grid
```
sschmidt23, line 114: As the pandas doc page for .attrs [points out](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.attrs.html) "attrs is experimental and may change without warning." which makes me a little wary of storing the grid this way, though I don't have many better alternatives.  As I mentioned in a previous Slack comment, I ran into an error due to `attrs` in an earlier version of Pandas, so we should add a >1.?? requirement to setup.py for Pandas compatibility.

sschmidt23, line 114: As the pandas doc page for .attrs points out "attrs is experimental and may change without warning." which makes me a little wary of storing the grid this way, though I don't have many better alternatives. As I mentioned in a previous Slack comment, I ran into an error due to attrs in an earlier version of Pandas, so we should add a >1.?? requirement to setup.py for Pandas compatibility.

sschmidt23 (changes requested):
Overall looks good, name changes requested by Alex do help clarify things, though I do think we should move flowEngine.py to an `engines` subdirectory for clarity.  On the subject of the flow (or other model): is there a simple way to carry around variables that we may expect little/no correlation that we carry about in the correlated data, but that we might want to use in degradation?  The concrete example I'm thinking of is RA/DEC and blending: I don't think we want any correlation between ugrizy+redshift and RA/DEC (to first order), but to emulate blending we might want to check for object that are within 1" and modify their ugrizy magnitudes during the degradation phase.  I think this works currently by carrying extra columns in the data frame that aren't in the trained flow, but I didn't actually test that it worked.  

There are also various minor linting issues that you can see in the flake8 section of the GitHub action that should be fixed, e.g.:
```
/rail/creation/__init__.py:3:1: F401 '.flowEngine.FlowEngine' imported but unused
19
./rail/creation/__init__.py:3:35: W292 no newline at end of file
./rail/creation/degradation/__init__.py:1:39: W292 no newline at end of file
./tests/creation/test_degraders.py:63:69: W292 no newline at end of file
```
(though note that there are several imports that are necessary but flake8 doesn't like but that are necessary, so a few of the messages are erroneous).

So, a couple minor things (move flowEngine to subdir, think about whether and alternative to attrs exists for the grid storage, add pandas >= in setup.py), but other than that I think this is ready to merge

rail/creation/creator.py
jfcrenshaw: Changes I made in response to this review:
1. moved the engines to an `engines` subdirectory
2. added pandas >= 1.1 requirement to setup.py
3. fixed the linting issues

On the additional comments above:

1. The flow can't produce variables it doesn't know anything about. We could add info like RA/DEC after sampling from the flow, but before performing degradation. However, I think it makes more sense to just train the flow on those variables as well. That way sampling from the flow will provide galaxies with a realistic distribution of all of the relevant variables. For example, here is a large flow trained on some DC2 data:

image

Using a flow like this would allow the creation module to provide any variables we want alongside redshift and magnitudes. (Note that corner plot was made by Alex Gagliano)

2. I agree about the problems with `df.attrs`. However, I think it's fine to leave as is for now, as we are planning to move pdf storage to `qp` anyway. However, if something like this is still needed, maybe we should switch to Astropy tables, which I think has a more robust way to store table metadata.

jfcrenshaw:
Just to consolidate some of the above discussions, I think the four big outstanding issues that need to be addressed in future pull requests are:
1. Building an error model into the creation branch, including marginalizing over missing bands.
2. Allow inclusion of a second model for ancillary data (e.g. RA and Dec)
3. Switching pdf storage to `qp`.
4. Unifying the output format of `Creation` with the input format of `Estimation`.

rail/creation/creator.py
sschmidt23: Ok, my main concern was that there might be some overfitting or weird spurious correlations if the flow fixated on some minor systematic in what should be uncorrelated data, but this was more of a hunch than something I thought a ton about.  Including RA/DEC in the model for eventual degradation is actually a good idea.

As an aside, good to see the correlations with physical params like SFR and logM, I had not tried that, and the bimodal service distribution as well!

jfcrenshaw: Actually I thought some more about this, and realized this is problematic with respect to the “true” photo-z posteriors. If you calculated redshift posteriors with this larger flow, you’d need to marginalize over all the other non-magnitude variables to get a photometry-only posterior. However, there is a straightforward way around this, involving a second conditional flow:
1. Draw redshift and magnitudes from the first flow, i.e.  z, mags ~ flow1
2. Draw RA, Dec, etc from a second (conditional) flow. i.e.  RA, Dec ~ flow2(z, mags)
3. Apply degradation
4. Calculate true photo-z posteriors using flow1

sschmidt23 (approved):

sschmidt23:
I agree with your main three remaining points, I think we can create issues for each of those and deal with them in more focused PRs after this gets merged in.  

I'm happy to see this merged, so I'll approve changes, not sure if @johannct  or @aimalz had any remaining comments or requested changes.

johannct:
Honestly, I had 0 time. I am happy to see it merged. We can always open other issues if needed, it is never good that a PR spend too much time being held, and it should always concern as limited an amount of coding as possible.

rail/creation/creator.py
aimalz: This might be a good thing to include in a demo, but it doesn't need to be done for this PR, so I'll mark the conversation as resolved.

PR #14, opened 2020-08-21
Issue/7/creation baseline

jfcrenshaw:
Finished merging the normalizing flow classes I built into the workflow set up by Bryce. Streamlined a little bit. Cleaned up the folder by removing old files.


rail/creation/realNVP.py
```diff
@@ -0,0 +1,297 @@
+# From Francois Lanusse
```
jbkalmbach, line 1: Is this different from what's in the original repo? Is there any reason not to just require Francois' repo as a dependency? We could add in a setup script to add the repo location to the path so it just works.

jbkalmbach (approved):
Everything looks good. Only comment is on how to handle Francois' code.

rail/creation/realNVP.py
jfcrenshaw: That file is the same as the one from Francois' repo. I just made that change so that it would be easily included in our repo, but I am happy to include his repo as a dependency in a future commit. For reference, here is the [link to Francois' repo](https://github.com/EiffL/jax-nf).

PR #16, opened 2020-09-04
Issue/13/parallel io

sschmidt23:
This may need some modification to be fully in line with what you have in mind, but I think this is essentially what you asked for.  @aimalz take a look and see if I should change anything.

### DIFF
--- a/rail/estimation/main.py
+++ b/rail/estimation/main.py
@@ -1,7 +1,7 @@
 import sys
 import yaml
 from estimator import Estimator as BaseEstimation
-from utils import base_yaml
+from utils import *
 import algos
 
 #Note: This is where 'base.yaml' actually belongs, but how to make it so 
@@ -31,7 +31,8 @@ def main(argv):
     
     pz.train()
 
-    for i, (start, end, data) in enumerate(pz.iter_chunk_hdf5_data(pz.testfile)):
+    for i, (start, end, data) in enumerate(iter_chunk_hdf5_data(pz.testfile,
+                                                                pz._chunk_size)):
         if i==0:
             outhandle = pz.initialize_writeout()
         pz.test_data = data
--- a/rail/estimation/main.py
+++ b/rail/estimation/main.py
@@ -31,10 +31,15 @@ def main(argv):
     
     pz.train()
 
-    pz.run_photoz()
+    for i, (start, end, data) in enumerate(pz.iter_chunk_hdf5_data(pz.testfile)):
+        if i==0:
+            outhandle = pz.initialize_writeout()
+        pz.test_data = data
+        pz.run_photoz()
+        pz.write_out_chunk(outhandle, start, end)
 
-    pz.write_out()
-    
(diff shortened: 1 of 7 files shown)


rail/estimation/main.py
```diff
@@ -31,11 +31,16 @@ def main(argv):
...
     pz.train()
 
-    pz.run_photoz()
-
-    pz.write_out()
+    outf = initialize_writeout(pz.saveloc, pz.num_rows, pz.nzbins)
     
+    for start, end, data in iter_chunk_hdf5_data(pz.testfile,pz._chunk_size):
+        pz.test_data = data
+        pz.run_photoz()
```
joezuntz: Is there a reason the `run_photoz` method doesn't return accept `test_data` and return `pz_dict`, rather than setting and getting attributes before and after?

rail/estimation/utils.py
```diff
@@ -23,35 +43,49 @@ def load_raw_hdf5_data(infile):
...
-    return data
+    data = {}
+    num_rows = get_input_data_size_hdf5(infile)
+    f = h5py.File(infile,"r")
+    for i in range(0,num_rows,chunk_size):
+        start = i
+        end = i+chunk_size
+        if end > num_rows:
+            end = num_rows
+        for key in f.keys():
```
joezuntz, line 74: Do you always want to read all keys here?

rail/estimation/main.py
```diff
@@ -1,7 +1,7 @@
 import sys
 import yaml
 from estimator import Estimator as BaseEstimation
-from utils import base_yaml
+from utils import *
```
joezuntz, line 4: `import *` is usually regarded as a bad idea as it's very opaque.

joezuntz (commented):
I think I understand the basic shape here, and it seems to make sense.  I've added a couple of comments / questions.

rail/estimation/utils.py
sschmidt23: The idea was to give people writing the subclass access to all possible columns rather than restricting to some subset.  For example, say someone writes a fancy code that uses positional information and wants to use RA/DEC.  If we have files with many extra columns that aren't used then we could modify this, but currently that isn't the case.

rail/estimation/main.py
sschmidt23: Alex mentioned a potential reorganization of utils.py, so I was waiting on that before doing anything more sophisticated, i.e. this import * will hopefully go away in a future update.

rail/estimation/main.py
sschmidt23: No, I had some reason I did it this way that I can't remember, but data can just as easily be fed in as an argument and the dictionary returned by run_photoz(). That really makes more sense now that we are only reading in a subset of the data each time.  I'll update this in the code.

sschmidt23: switched data to a function input and pz_dict as a return value for run_photoz in newest commit

PR #19, opened 2020-09-15
Issue/12/est tests

johannct:
pulling from Sam's branch to allow for comparison and discussion


johannct:
ok I committed and merged to be able to test the parallel_io contained in Sam's branch..... but the test files are not accessible at NERSC.... We need to solve three issues to move forward : 
 - base.yaml cannot lay around the package as it does now, and basic API quality requires that it be passed as argument to estimator init (my 2 cents)
 - test files need to be either included in RAIL/examples if this is meant to stay, or included in a rail/test_data dir, or downloadable/installable at first call
 - I think rail needs to be lazier : loading the data as soon as the estimator is created does not seem like a good idea in the long run

PR #23, opened 2020-09-22
fix a bug in the action for coverage and add the test badge to README

johannct:
quick fix to coverage and esthetic addendum to README

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
     - name: Test with pytest
       run: |
-        pytest --cov=./ --cov-report=xml
+        pytest --cov=./rail --cov-report=xml
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1
       with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ name: rail
 
 on:
   push:
-    branches: [issue/3/estimation-template, $default-branch ]
+    branches: [issue/3/estimation-template, issue/2/ci, $default-branch ]
   pull_request:
     branches: [issue/3/estimation-template, $default-branch ]
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,9 +5,9 @@ name: rail
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
     - name: Test with pytest
       run: |
-        pytest --cov=./ --cov-report=xml
+        pytest --cov=./rail --cov-report=xml
```
sschmidt23: Since the tests now live in the directory above the rail package, I think specifying `./rail` rather than `./` means that the tests are not found and do not run, as you can see from the 0.0% coverage when I added the branch to the GitHub action to test that things worked properly.
So, I think we want to stick with `--cov=./`

sschmidt23: Changing it back to did, indeed, get the tests to run.

README.md
```diff
@@ -1,3 +1,4 @@
+![tests](https://github.com/LSSTDESC/BlendingToolKit/workflows/tests/badge.svg)
```
sschmidt23, line 1: Badge is working properly.

johannct:
hmmm @sschmidt23 ok I see your point, but if you look at BlendingToolKit it has the same organisation, with the `btk` package at the same level as the `tests` level, and still the yaml file has `--cov=./btk` : https://github.com/LSSTDESC/BlendingToolKit/blob/master/.github/workflows/config.yml

johannct:
so there may be something else to understand..... but clearly the coverage report is dysfunctional for rail.

sschmidt23 (changes requested):
As noted in comments, specifying --cov=./rail skips the tests that live in the RAIL directory, so I think we want to keep --cov=./ in main.yml.  The coverage badge is working properly.  Before merging, we should remove the issue/2/ci branch from the on: push: in main.yml (I need to look up how to trigger on the current branch and potentially add that in its place).

.github/workflows/main.yml
sschmidt23: Johann pointed out that the coverage was incorrect: it was counting the test files but not the actual rail files.  This is fixed by running pytest as a package.  I pushed this change and coverage now appears to be functioning properly

johannct (commented):
ok let's try that on master

johannct (commented):
remove afterwards

sschmidt23 (approved):

PR #26, opened 2020-09-24
Issue/22/class name

### LINKED ISSUE #22: define class_name in config files so that same code with different parameters  can be run concurrently

sschmidt23:
This PR addresses issue #22 to read the class name from the config file.  It also adds functionality to perform multiple code runs from a single config file, which necessitates moving the pz.saveloc outside of the class and back to main.py.  I also added a nonexistent.yaml to test the case where the class name is not actually present in the set of algos.

some of the names in the new "runs" structure of the yaml may not be what everyone will prefer, same with the output filename structure, so those are definitely up for comment/revision.


examples/configs/trytworuns.yaml
```diff
@@ -0,0 +1,17 @@
+runs:
+  run1:
+    run_params:
```
johannct: @sschmidt23 why do we still need the `run_params` level here? Do you foresee something else under run1 but outside of run_params?

examples/main.py
```diff
@@ -18,34 +18,42 @@ def main(argv):
...
-        pz_dict = pz.estimate(data)
-        write_out_chunk(outf, pz_dict, start, end)
-        print("finished " + name)
-
-    finalize_writeout(outf, pz.zgrid)
+        overall_dict = yaml.safe_load(f)
+
+    for single_run_dict in overall_dict['runs'].values():
+
+        print(single_run_dict)
```
johannct: remove the print statement

examples/main.py
```diff
@@ -18,34 +18,42 @@ def main(argv):
...
-    for start, end, data in iter_chunk_hdf5_data(pz.testfile, pz._chunk_size,
-                                                 'photometry'):
-        pz_dict = pz.estimate(data)
-        write_out_chunk(outf, pz_dict, start, end)
-        print("finished " + name)
-
-    finalize_writeout(outf, pz.zgrid)
+        overall_dict = yaml.safe_load(f)
+
+    for single_run_dict in overall_dict['runs'].values():
```
johannct: Brainstorming aloud here : this is ok, but we loose access to the fields `run1` `run2`. I think it is perfectly fine to leave these arbitrary : they are user choices to separate the different config..... But then they also might quickly end up being thought of as identifier strings for each config, and thus we may want to build a mechanism to attach these to their configs, and actually persist the field? I realize that there is an outfile_tag as well, but this translates into a fiilename I presume, and I dislike coding relevant stuff in filenames unless forced.

examples/main.py
```diff
@@ -18,34 +18,42 @@ def main(argv):
...
+        print(f"code name: {code}")
+
+        pz = code(base_config, single_run_dict)
+        pz.inform()
+
+        if 'outfile_tag' in single_run_dict['run_params']:
+            outfile = f"{name}_{single_run_dict['run_params']['outfile_tag']}.hdf5"
+        else:
+            outfile = f"{name}.hdf5"
+        print(outfile)
```
johannct: remove print statement

johannct (approved):
LGTM : there are a couple of print statements to remove and 2 broader scope comments that may be relevant for this pull request, but I approve it so that there is no block to merge.

examples/configs/trytworuns.yaml
aimalz: I assume it's for running multiple test conditions for a single code using one config file.  I think we might want to force one config file per run, though, because each Estimator object corresponds to one config file rather than one run right now, and introducing a mandatory `run#` field seems clunky when we probably won't be using it for the real experiments.

aimalz (changes requested):
I'm in favor of resolving Johann's comment about parsing the config files with mandatory `run#` by eliminating that layer.  Could the `code_name`/`class_name` distinction instead be reframed as `algo_name`/`run_id`?

examples/main.py
sschmidt23: removed

examples/main.py
sschmidt23: removed

examples/configs/trytworuns.yaml
sschmidt23: So based on the discussion this morning, I removed the ability to do multiple runs in a single config and essentially reverted back to the way we had things before.  But, now with the addition of an "output_dir" directory name that is prepended to the output path before the output filename.  That way you can save multiple versions of a code's output without overwriting.  
Is that what we had in mind for the structure?

sschmidt23: Ah, just as I push this, I realize that this could be better organized as Alex stated below, with a directory named for the code and then output files based on the run_id, which does then keep like codes together, let me switch to that instead.

sschmidt23: Ok, new setup has a 'run_name' parameter in the yaml file and creates [outpath]/[code_name]/[run_name].hdf5 output files for each run

rail/estimation/utils.py
```diff
@@ -92,7 +92,7 @@ def iter_chunk_hdf5_data(infile, chunk_size=100_000, groupname='None'):
 def initialize_writeout(outfile, num_rows, num_zbins):
     outdir = os.path.dirname(outfile)
     if not os.path.exists(outdir):
-        os.mkdir(outdir)
+        os.makedirs(outdir)
```
johannct: Do we want to keep the default `exist_ok=False` for `makedirs`. I think that existence of the directories should not raise an error in this context. See https://docs.python.org/3/library/os.html

johannct (changes requested):
Do we want to keep the default exist_ok=False for makedirs in utils.py. I think that existence of the directories should not raise an error in this context. See https://docs.python.org/3/library/os.html

rail/estimation/utils.py
sschmidt23: I tested this on my machine (cori is still down so I could not test there), and it did not raise an exception without the flag.  But, you're right, I'll add the flag just to be sure.

johannct (approved):

johannct (approved):
LGTM

aimalz (approved):

Issue #27, opened 2020-10-01
Linting and tests


sschmidt23:
Alex mentioned many remaining problems with linting in estimation code, need to fix this up, as well as add some remaining tests for a few remaining lines in estimation and utils

aimalz:
closed by #28

PR #28, opened 2020-10-02
Issue/27/linting

sschmidt23:
This addressees issue #27, fixing some linting issues and adding a couple more tests and test files.

One question: the main test/validate files have 10k and 20k DC2 galaxies, included so that you get fairly decent results with e.g. FlexZBoost when you train on the training file and produce photo-z's for the validation.  But, the tests take approx. 5 mins to run on GitHub actions, almost entirely due to FlexZBoost and its training.  Should we reduce the numbers of galaxies in the test/validate files to speed things up, or is 5 mins acceptable for checking the results with a reasonable input set of data?

Also, if using pandas h5 or parquet files you need to include tables and yarrow respectively.   I did not put these in to the setup.py file, I think we were still discussing whether to use an extras_requires or some other option I didn't remember for loading dependencies.  So, that is left for a future PR.


aimalz:
Re: time, I agree that we should reserve GitHub actions for unit testing.  I think this is what @johannct was getting at with #29, so it doesn't necessarily need to be resolved for this PR, but it might speed things up if we address it soon.

Re: file formats, I think @eacharles will want to weigh in on this, so let's consider what we have to be a placeholder until then.

sschmidt23:
The last push, [00afcdf](https://github.com/LSSTDESC/RAIL/pull/28/commits/00afcdfbc33859006087aba793a50a7bdbf3ff90), adds small files with 100 galaxies for training and 10 for validation and adds tests that the 10 `zmode` values are reproduced for the three current codes, which addresses issue #29.  The larger 10k and 20k galaxy files will be kept on for a demo notebook mentioned in issue #30, which I'll get to in a near future PR.

And yes, file formats were always meant to be placeholder until the qp development is finished.

tests/test.yaml
```diff
@@ -0,0 +1,7 @@
+base_config:
+  trainfile: tests/data/training_100gal.hdf5
+  testfile: tests/data/validation_10gal.hdf5
+  hdf5_groupname: photometry
+  chunk_size: 1000
+  configpath: ./configs
+  outpath: ./results
```
aimalz, line 7: Looks like one little linting error got through, ha!

aimalz (approved):
This is solid!

Issue #18, opened 2020-09-14
Define the mechanism to discover and check presence of usable PZ estimation codes


johannct:
Currently, estimation.py includes an attempt to look into the content of the algos dir, based on the name of local config files. This is clearly fragile and very rigid, and was not meant as a perennial solution. We need to provide for the fact that : 
 - users may not wish to install all the possible codes that RAIL interfaces; if too complicated rail needs to clearly require external packages
 - users may wish to subselect a list of codes from all the ones they have isntalled for a given run.

johannct:
There is certainly a standard install procedure for the first point : 
`    extras_require={
        'flexcode':  ["flexcode[all]"],
    }
)`
in setup.py allows one to do `pip install .[flexcode] --user` and this will do the same as `python setup.py install --user` but downloading and installing `flexcode` and `xgdboost` in the process. The problem I can imagine in this very simple option is that 1/ one likely does not have much flexibility in how to install all these, and 2/ this does not allow for custom build of rail depending on which  packages are called.

johannct:
I think we defined a track to solve this, so closing

PR #31, opened 2020-10-02
Move to a more standard way to instantiate algorithms from a Factory

johannct:
We depend currently on a cascade of import in order to trigger under the hood the construction of a dictionary of all the available algorithm classes. Instead we should try : 
- to turn it into a more conventional factory logic
- to avoid building a complete dictionary of available algorithms as soon as the package is imported or any single algorithm is requested

This pull request aims to solve #18


rail/estimation/estimator.py
```diff
@@ -19,6 +19,16 @@ class Estimator(object):
 
     @classmethod
     def _find_subclass(cls, name):
+
```
johannct: the issue here is that `name` is the name of an algorithm class, not the name of the module file where it can be found. Hence the need to loop over all the files under algos.....

rail/estimation/estimator.py
```diff
@@ -19,6 +19,16 @@ class Estimator(object):
...
+        if name not in cls._subclasses.keys():
+            root_dir = os.path.dirname(__file__)
+            names = []
+            for filename in os.listdir(root_dir+'/algos'):
+                if filename.endswith('.py') and filename != '__init__.py':
+                    name = filename[:-3]
+                    print(name)
+                    importlib.import_module('.'+name,'rail.estimation.algos')
```
johannct: this line triggers __init_subclass__ below which loads the class object into the directory..... Hence all the available algorithms end up checked and loaded, which is not ideal.

rail/estimation/estimator.py
```diff
@@ -19,8 +20,23 @@ class Estimator(object):
 
     @classmethod
     def _find_subclass(cls, name):
```
sschmidt23, line 22: This is fairly straightforward when you work your way through, but with several levels of for, if, and with statements, this should probably have a docstring explaining what is happening with ast and importlib.

sschmidt23 (commented):
The functionality seems to work, loading only the single needed subclass when running e.g. the main.py script in examples, which could save some time when more codes are added with their various dependencies not needing to be loaded when a single algorithm is run.  I did leave a comment about _find_subclass needing a docstring, given the six levels of indent happening there.
I'll also leave this as a comment rather than an explicit approve until Eric or Alex takes a look to see that they approve of how you've implemented this.

rail/estimation/estimator.py
johannct: docstring committed and pushed by 935fabd . Is it ok?
I leave @eacharles and @aimalz to decide if we should rename the method, like `factory`. Also @aimalz expressed, if I understood well, the wish to also build an instance registry, which would need some work to build a second dictionary based on a key that would append a unique instance identifier to the class name of the instance..... This would probably be straightforward to add to `__init_subclasses__ as it has access to the run config dictionary.

sschmidt23: Yes, docstring looks good to me

rail/estimation/estimator.py
johannct: solution is to use abstract class semantics, as done in this pull request

aimalz (approved):
This makes total sense and is, I think, what we need for running a big batch script for lots of codes operating on a single set of inputs in a real experiment. I'm happy to ignore the small drop in code coverage statistics to get this merged in sooner.

johannct:
merging to have a clean codebase for the sprint

PR #37, opened 2020-12-03
Issue/36/biased flow

jfcrenshaw:
Added a notebook using creators to generate training and test sets for use in the estimation module. 

- test set -- unbiased sample from the normalizing flow saved in demo_flow.pkl.
          Saved in two files:
                    - testset_public.pkl -- dataframe containing galaxy magnitudes
                    - testset_private.pkl -- dataframe containing the corresponding true redshifts and posteriors

- training set -- biased sample, generated by drawing a sample from the unbiased normalizing flow, then applying a selection function in the creator object. These magnitudes and true redshifts are saved in a dataframe in trainingset.pkl.


aimalz (approved):
Looks great!

PR #40, opened 2020-12-04
Johannct/estimation/train z

johannct:
the trainZ pathological algorithm from the DC1 paper

### DIFF
--- a/examples/main.py
+++ b/examples/main.py
@@ -10,11 +10,11 @@ from rail.estimation.estimator import Estimator
 def main(argv):
     if len(argv) == 2:
         # this is in case hiding the base yaml is wanted
-        base_config = 'base.yaml'
         input_yaml = argv[1]
+        base_config = 'base.yaml'
     elif len(argv) == 3:
-        input_yaml = argv[2]
-        base_config = argv[1]
+        input_yaml = argv[1]
+        base_config = argv[2]
     else:
         print(len(argv))
         print("Usage: main <config yaml file> [base config yaml]")
--- a/examples/main.py
+++ b/examples/main.py
@@ -13,8 +13,8 @@ def main(argv):
         base_config = 'base.yaml'
         input_yaml = argv[1]
     elif len(argv) == 3:
-        input_yaml = argv[1]
-        base_config = argv[2]
+        input_yaml = argv[2]
+        base_config = argv[1]
     else:
         print(len(argv))
         print("Usage: main <config yaml file> [base config yaml]")
--- a/rail/estimation/algos/trainZ.py
+++ b/rail/estimation/algos/trainZ.py
@@ -9,6 +9,7 @@ N (z) of the training set.
 import numpy as np
 from rail.estimation.estimator import Estimator as BaseEstimation
 
+
 class trainZ(BaseEstimation):
 
     def __init__(self, base_config, config_dict):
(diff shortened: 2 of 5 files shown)


rail/estimation/algos/trainZ.py
```diff
@@ -0,0 +1,41 @@
...
+        self.cdf = cdf / cdf[-1]
+        self.zgrid = self.midpoints
+        np.random.seed(87)
+
+    def inform(self):
+        pass
+
+    def estimate(self, test_data):
+        test_size = len(test_data['id'])
+        zmode = np.repeat(self.zmode, test_size)
```
sschmidt23, line 38: You originally had a random draw here, but we are currently predicting the mode of the PDF not a sample from it, so I replaced with the mode.  If we add samples at a later date then the old code could be added back.

tests/test_algos.py
```diff
@@ -59,3 +59,15 @@ def test_flexzboost():
     pz_algo = flexzboost.FZBoost
     pz_dict = one_algo(pz_algo, config_dict)
     assert np.isclose(pz_dict['zmode'], zb_expected).all()
+
+
+def test_train_pz():
+    config_dict = {'run_params': {'zmin': 0.0,
+                                  'zmax': 3.0, 'nzbins': 301}}
+    zb_expected = np.repeat(0.1445183, 10)
```
sschmidt23, line 67: I updated to the modes here so that the test would pass

rail/estimation/algos/trainZ.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, base_config, config_dict):
+        super().__init__(base_config=base_config, config_dict=config_dict)
+
+        inputs = self.config_dict['run_params']
+        self.zmin = inputs['zmin']
+        self.zmax = inputs['zmax']
+        self.nzbins = inputs['nzbins']
+
+        zbins = np.linspace(self.zmin, self.zmax, self.nzbins+1)
```
sschmidt23, line 23: You had a bug here where the array sizes did not match up, which caused an error when running the code.  The histogram centers are really what we care about so I changed things so that we have nzbins number of bin centers and spit those out (so we might want to change the default from 301 to 300)

examples/main.py
```diff
@@ -13,8 +13,8 @@ def main(argv):
         base_config = 'base.yaml'
         input_yaml = argv[1]
     elif len(argv) == 3:
-        input_yaml = argv[1]
-        base_config = argv[2]
+        input_yaml = argv[2]
+        base_config = argv[1]
```
sschmidt23: Why did you change this back?  I thought you were the one who wanted the order swapped so that if you do e.g.
`python main.py configs/trainZ.yaml` it assumes base.yaml for the base_config, and to run with a different base config it would be e.g.
`python main.py configs/trainZ.yaml newbaseconfig.yaml`, i.e. the algorithm config does not change from first to second positional argument.

sschmidt23 (commented):
Looks good, I fixed small linting errors, and also changed from random samples to actually spitting out the mode for zmode.  One question remains about the argument ordering for main.py, and there might be one or two more remaining linting issues if you run flake8 (extra imports I think) that you might want to take care of before we merge.

rail/estimation/algos/trainZ.py
johannct: correct, sorry about that

examples/main.py
johannct: mistake! I helped Julia fix her running of rail and for some reason she had the old order, so I reversed mine to check that this was indeed her problem..... and forgot to reverse back :(

johannct:
> Looks good, I fixed small linting errors, and also changed from random samples to actually spitting out the mode for zmode. One question remains about the argument ordering for main.py, and there might be one or two more remaining linting issues if you run flake8 (extra imports I think) that you might want to take care of before we merge.

main.py was a commit error, and I only fixed the BaseEstimation unused import..... I think that we need the other one, despite flake8 warning.

sschmidt23 (approved):
Yes, looks good now, I think it's ok to merge

PR #44, opened 2021-01-05
Add new init mechanism and test

joezuntz:
This small change allows users to initialize estimators from a dictionary directly as well as a file.


johannct (approved):
LGTM. We did not implement it because it will not be the standard way to instantiate it, but it is fine to keep it as an option I guess. @aimalz ?

aimalz (approved):

Issue #45, opened 2021-01-05
Running the inform and estimate stages in different processes


joezuntz:
Hi all,

Because the training data used by `inform` method will typically be of a relatively small size and run on a single node, whereas the estimation can (optionally) be done in parallel by many nodes, it would be useful to be able to split these two phases.

This could be done just by ensuring that future Estimator subclasses, like the ones that are currently there, can be pickled.

As an addendum - is the intention that the testfile specified in the estimator is used within the library to test algorithm performance?  Or is it intended to be used for the eventual full catalogs?  I think it's the former, but if not then it would also be useful to remove the requirement that this be specified in the init method.

Does this sound okay?  I can add a picklability test if so.

joezuntz:
For example (copying this from an email for clarity), I'd like to do something like:

```
# in first (training) process
cls = Estimator._find_subclass("FZBoost")
pz = cls(base_config, run_dict)
pz.inform()
with open('model.pickle', 'wb') as f:
    pickle.dump(pz, f)
# finish process
```

Then in another process, later:

```
with open('model.pickle', 'rb') as f:
    pz = pickle.load(f)
 
for data in iter_chunk_hdf5_data(cat_file, chunk_size, name):
    pz_dict = pz.estimate(data)
    # save pz_dict data somewhere
```

sschmidt23:
I think that there are two separate use cases, the two that you mention in particular, though @aimalz can say what he had in mind for sure.  It is currently coded for something like a data challenge, where we are testing all algorithms with a single training file and a single test file.  It is relatively straightforward to modify the set up to read in some form of pickled result of the training for each code for the second case where we want to generate large catalogs, as you say we just read in the trained model rather than running the inform stage.  The form will be different for different codes, so we may want to think a bit about whether any solution is generic enough to cover all use cases (it won't always be just a pickle file, in some cases it will be a set of parameters), though I believe it may just end up "hardcoded" in the estimate function for each subclass as to what exactly is read in.  I think we started to discuss this at some point, and one reason that I didn't implement it was because we were still discussing provenance tracking, and how to keep track of the setup used to create such a pickle file.

johannct:
Ok it does sound like something we definitely need to think over and likely pursue, but it needs some thoughts and trial and error stage regarding implementation, based on a sufficiently complex use case. @joezuntz Do you have such a use case at hand? I think within TXPIPE you also gave thought to provenance..... anything that can be pulled over and used more globally?

[...]

PR #46, opened 2021-01-13
Issue/45/inform option 

sschmidt23:
This simple PR addresses one isolated portion of issue #45 (the main one, but discussion in that issue grew beyond the specific topic).  It adds a new `inform_options` dictionary to the yaml config file that control whether to write out a trained model or read in a trained model if a pre-trained model exists.  This PR does *not* address the provenance concerns that come with creating/ingesting pre-trained models, that will be left for a future update following discussions of what provenance tracking system that we want to use.

I added tests that create a model and then compare the outputs when ingesting that model vs an initial run, so it looks like everything functions as it should.


.github/workflows/main.yml
```diff
@@ -5,7 +5,7 @@ name: rail
 
 on:
   push:
-    branches: [ master ]
+    branches: [ issue/45/inform_option ]
```
johannct: change that on final push of course

examples/main.py
```diff
@@ -34,7 +34,14 @@ def main(argv):
...
 
     pz = code(base_config, run_dict)
-    pz.inform()
+
+    pz.inform_dict = run_dict['run_params']['inform_options']
+    if pz.inform_dict['load_model']:
+        # note: specific options set in subclasss func def
+        pz.load_pretrained_model()
+    else:
+        pz.inform()
```
johannct, line 43: Can we have a little doc somewhere (README of estimation?) explaining the logic : I was expecting the contrary (run_model true by defualt, and if set to false then load model). It is arbitrary of course, so just a doc to explain the API decision would be useful. This is where one can appreciate the power of C enums, and what a pain it is that python does not have a clean mechanism for it

rail/estimation/estimator.py
```diff
@@ -64,6 +64,13 @@ def inform(self):
         """
         raise NotImplementedError
 
+    def load_pretrained_model(self):
+        """
+        If inform step has been run separately, this funciton will
+        load the information required to run estimate
+        """
+        raise NotImplementedError  # pragma: no cover
+
```
johannct, line 99: Sam, I am not sure about that.... The code you put in flexzboost and sklearn_nn seems rather boilerplate, so why not put it here, and overwrite with a pass randomPZ as you did, when it is not applicable. In the end all these code need to actually move out of estimation actually, as we want to centralize IO; but that may be left for a later stage

tests/test_algos.py
```diff
@@ -9,6 +10,10 @@
 def one_algo(single_estimator, single_input):
     """
     A basic test of running an estimator subclass
+    Run inform, write temporary trained model to
+    'tempmodelfile.tmp', run photo-z algorithm.
+    Then, load temp modelfile and re-run, return
+    both datasets.
```
johannct, line 17: I wonder if it would not be better to have a dedicated test for loading, for saving, etc.... I understand it is painful, if you want I can implement it to finsih the review. I am afraid that quickly the test calls become too complex to easily point to regression issues in the future.

rail/estimation/algos/flexzboost.py
```diff
@@ -141,6 +143,14 @@ def inform(self):
                 bestsharp = sharp
         model.sharpen_alpha = bestsharp
         self.model = model
+        if self.inform_options['save_train']:
```
johannct, line 154: we need some boilerplate here, unfortunately, to check early (before starting to train) that modelfile is provided if save_train or load_train is True. I just experienced wasting training time by ending with:
```
finding best sharpen parameter...
Traceback (most recent call last):
  File "main.py", line 67, in <module>
    main(sys.argv)
  File "main.py", line 43, in main
    pz.inform()
  File "/home/cohen/lsst/codes/pipelines/photoz/RAIL/rail/estimation/algos/flexzboost.py", line 147, in inform
    with open(self.inform_options['modelfile'], 'wb') as f:
KeyError: 'modelfile'
```

rail/estimation/algos/flexzboost.py
```diff
@@ -141,6 +143,14 @@ def inform(self):
...
+            with open(self.inform_options['modelfile'], 'wb') as f:
+                pickle.dump(file=f, obj=model,
+                            protocol=pickle.HIGHEST_PROTOCOL)
+
+    def load_pretrained_model(self):
```
johannct: what code is in charge of calling `load_pretrained_model`, and based on which set of parameters? In the tests we probably need to exercise : 1/ train and stop (no estimate); 2/ train and estimate; 3/ load and estimate;

johannct (changes requested):
thanks for this Sam, let's try to have a quick turnaround on my comments. I am available to pair code if we find a good time (though next week is going to be tough)

examples/main.py
sschmidt23: I think a jupyter notebook that will be added with issue #51 will take care of this, having examples written out in a notebook I think will suffice, but I'll say lets leave it for that separate issue.

tests/test_algos.py
sschmidt23: Separate tests for loading and saving would be good, this test was more for actually checking that the results from loading a trained model agree with the results from computing the model.  Were you thinking of e.g tests where self.modelfile is not defined and such?

.github/workflows/main.yml
sschmidt23: will fix this once the PR is accepted

rail/estimation/estimator.py
sschmidt23: I moved the 'boilerplate' to estimator.py and put a note that subclasses can overwrite if they are loading something other than a pickle file

rail/estimation/algos/flexzboost.py
sschmidt23: I added a check in the __init__ for the two codes that write out a model, and similarly put a check on the modelfile in estimator.py in load_pretrained_model to check that the `modelfile` keyword is there and that the file exists.  These are not currently covered by the tests, which is related to your comment below about separate tests, I think.

rail/estimation/algos/flexzboost.py
sschmidt23: I was reworking this a bit in separate branch, but load_pretrained_model is handled in main.py currently, and yes, separate tests are probably necessary.  I wasn't thinking of "train and stop" as an option, but I guess that is a potential use case.

rail/estimation/algos/flexzboost.py
sschmidt23: We can add tests of coverage once you confirm that this answers your initial comment

.github/workflows/main.yml
johannct: sure

examples/main.py
johannct: ok can you put a note there?

tests/test_algos.py
johannct: I was thinking of all the cases where codecov notices lack of coverage : try/except blocks on loading and saving pretrained estimators

johannct (approved):
LGTM, just these items : 
- add a note to #51 as a little doc on how to pretrain and save
- create a little test code to exercize saving/loading of pretrained and the failure cases caught by try/except conditions in the code

tests/test_algos.py
sschmidt23: I just added tests that I think cover all the examples of saving and loading (the one_algo function saves a model, then loads the model and compares the two outputs, so save/load is tested in the course of that), plus tests checking the try/except for missing keys and files that weren't covered before.  So, I think we're now good.

Can you take a quick look and see if there are any cases that I've missed?  codecov is happy, but I'm trying to think if there might be an additional missing use case.

rail/estimation/algos/flexzboost.py
sschmidt23: coverage tests were added so I'll resolve this

sschmidt23:
I added a note on issue #51 and added some new tests.  @johannct can you take one quick look at the tests to see if I've covered everything from a user point of view, I think I have but a quick eye on the code to be sure is good.  Once you sign off I'll resolve the merge conflict and merge.

tests/test_algos.py
johannct: LGTM thanks Sam. Remember to modify the action as well (to push on master)

Issue #34, opened 2020-12-01
Autodoc to formalize documentation


aimalz:
The title says it all! ~~I'm going to steal the sphinx setup from [CLMM](https://github.com/LSSTDESC/CLMM) as a template to get started, but anyone else is welcome to jump in on this.~~ Yikes, that's got more bells and whistles than we need at this stage. @sjoudaki will be taking it from here and starting from something simpler like [qp](https://github.com/aimalz/qp).

johannct:
I add here that when ready to tag RAIL, autodoc needs to be ready, but also README.md will need to be updated, especially with detailed mentions of how to build and what are the different required and optional packages. Pinging @sschmidt23 on this

aimalz:
#82/#83 address the final comment here, so I'm going to close this issue in light of #53.

PR #53, opened 2021-03-05
Debugging autodoc

aimalz:
[This seems to work now.](https://lsstdescrail.readthedocs.io/en/issue-34-autodoc-debug/) I started from the last working version of the branch for #34 and just removed the locally generated .rst files (though I think I may have missed some) and it just worked. I guess I broke some things fiddling with the various configs before fixing what we suspect was the real problem in the settings on readthedocs.


sjoudaki:
Hi Alex,

When I click on the link you provide and then on "RAIL", I find:

Screen Shot 2021-03-05 at 7 03 48 PM

I even tried to clean the cache completely, but to no avail. When I then go into the builds and click on the latest one that passed (#13171658) and then on "view docs", I find:

Screen Shot 2021-03-05 at 7 04 17 PM

So, not sure if this is due to my laptop or still not working?

sjoudaki (approved):
Will try to merge the branch issue/34/autodoc.

Issue #4, opened 2020-01-09
Evaluation module boilerplate


aimalz:
[The PZ DC1 metrics code](https://github.com/LSSTDESC/PZDC1paper/tree/master/metric_scripts) can be used as a starting point for the submodules that evaluate the accuracy of photo-z PDF estimates.  Note that many of those functions were built on [qp](https://github.com/aimalz/qp), which is at this point pretty old and could use an upgrade, which will occur in [a DESC fork](https://github.com/LSSTDESC/qp).

Steps (Julia):  (starting at sprint week 2020)

- [x] Clone the RAIL repository and create a new branch (follow the branch name convention).

- [x] Quick review of PZ DC1 paper.

- [x] Study the current implementation available in the PZ DC1 metrics code. 

- [x] Find toy data to run quick tests while coding. 

- [x] Run FlexZBoost on the toy data to get input PDFs. 

- [x] Create a superclass for the generic concept of metric. It will be the mother of a number of classes, one for each metric. It should receive PDFs from any photo-z algorithm (standardized format), and allows the comparison between results from several codes. 

- [x] Create the metric classes (start with one PDF metric, e.g. the CDE loss, and run tests before going to the full list of metrics and point estimate reduction). 
 

- [x] Create a demo notebook showing the metrics evaluation using the new classes.

- [x] Run using the exact same data used in the DC1 paper to allow the reproduction of the paper's results (numbers in tables 2 and 3 from the paper must match). 

- [ ]  Write unit tests for the new classes.
- [ ] Update repo's documentation file.

- [ ] Open a pull request.

gschwend:
progress report: I had trouble to run FlexZBoost, on nersc, due to conda env issues. I am getting help from Johann Cohen-Tanugi via slack. We agreed that the best approach for now is to work locally in my laptop, which also is giving us a hard time to setup rail's dependencies (but we are almost there).

gschwend:
Finally, I can FlexZBoost on the test sample successfully. 

The requirements to run on my laptop were: 
```
$ conda create -n py37 python=3.7 pyyaml pandas 
$ conda activate py37
$ pip install flexcode[all]
$ conda install -c conda-forge openmp
$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
$ brew install libomp
 ``` 
Maybe it is a good idea to update the setup.py file to include all the dependencies.

sschmidt23:
[...]

PR #54, opened 2021-03-05
Issue/4/evaluation baseline

gschwend:
The work to address issue #4 is still ongoing, but I reached a certain point where I am stuck. Therefore, I am opening this draft pull request to ask for Sam's help, so he can make comments directly on the code if necessary. 

@sschmidt23 please start by looking at the notebook exported to HTML ( dc1_check_incomplete_for_revision.html) in the evaluation directory, where I discuss the metrics issues and show some relevant blocks of code (pasted as markdown, just for reference).  We can talk via zoom if you'd like me to give more context and show the details.  
Thank you in advance.


rail/evaluation/metrics.py
```diff
@@ -0,0 +1,376 @@
...
+
+    Parameters
+    ----------
+    metrics: `metrics` object
+        instance of metrics base class which is connected to a given sample
+    """
+
+    def __init__(self, metrics, scipy=False):
+        if scipy:
+            cvm_result = stats.cramervonmises(metrics._pit_dist, "uniform")
```
sschmidt23: The scipy.stats.cramervonmises takes *samples* for the first argument, not a histogram of values, so metrics._pit_dist should instead be metrics._pit.  If you make this change then I get exact agreement with skgof for the value of the CvM for a given set of PIT values.

sschmidt23: Also cramervonmises is a very new addition to scipy, not present until v1.6.0, so we should update the setup.py requirements to scipy>=1.6.0

sschmidt23:
Before getting into the actual structure of the code, I started by taking a look at agreement with DC1 by running RAIL/evaluation on the FlexZBoost dataset on cori.  We saved the individual PIT values for each galaxy, so I checked agreement galaxy by galaxy on the PIT values in addition to looking at the distributions.  Comparing the values, they are very close in most cases, but some of the PIT values are off by a small amount <~0.02.  This is likely due to some differences in the assumed integration from the "old" qp to "new" qp (I recall old qp having some subsampling with a dz of a very small 0.001).  Attached is a plot of RAIL PIT - original PIT vs original PIT showing the residual between the old and new:
PIT_residuals

I am not super concerned with these small differences, though it would be nice to track them down by looking at what changed in the qp integration/cdf code.  

I checked the `scipy='True'` versions of  metrics.py KS and CvM against skgof and got absolutely identical answers to floating point precision when comparing the PIT values.  I found a bug in line 297 of metrics.py where you were comparing the histogram of PIT values rather than the array of PIT values in the scipy cramervonmises test, when you make that change things now work.  The values of KS and CvM do not exactly match DC1, but this is because the PIT values are slightly different.  However, if you were to feed in the original PIT values you would exactly match the value in the paper.
Anderson-Darling seems to be a bit of a mess: as 'uniform' test is not available in scipy we can't use that, and the fact that the metric diverges at the extrema means that it's not the most stable metric to compute, particularly in the presence of slight deviations in PIT values for our many many galaxies near the PIT extrema of 0 and 1.  Checking AD is going to be more of a pain.

For the `scipy='False'` versions of KS and CvM, I see a small difference when comparing to scipy and skgof for KS.  I think this is due to the fact that you are checkin the KS values on the n_quant grid rather than constructing the full empirical CDF for the almost 400k values.  For CvM the difference is much larger and I think we'll have to look in more detail at the implementation to track down the bug.

gschwend:
Screen Shot 2021-03-15 at 20 38 41
@sschmidt23 the screenshot shows that the update of qp brought significant changes to the PIT-QQ plot (using old qp on the right, using new qp on center). Now it is much more alike the plot shown in the paper (on the left).

gschwend:
Hi @sschmidt23, this week I racked my brain trying to reproduce DC1 results, still without success. I implemented a "debug mode" where the metrics are calculated using the original PIT values from DC1, as you did above. I haven't managed to achieve the same results, even when using scipy.stats tests as you did. I uploaded a compiled version of the notebook to show you my results and questions in the last commit.  About your statement in the comment above:  "_checkin the KS values on the n_quant grid rather than constructing the full empirical CDF for the almost 400k values_", I don't understand it. Could you explain a bit more, please? We can discuss on zoom if you prefer. Thanks a lot for the huge help.

gschwend:
All the issues mentioned in my previous comment were solved. Now I'm working on specific improvements suggested by the group during regular meetings.

johannct:
@Gschwend This PR is already quite long. I think you should consider it as the solution to one specific request or issue, and let the PR be reviewed and closed before improving the current state of the code. If I am not mistaken the idea here is to deliver a proper 'evaluation baseline', not the final codebase that we will carry all the way to operations :)
Stated otherwise, once you open the PR you should have to implement requests from the reviewers only. As far as I can tell, there is one comment from @sschmidt23 but it is not a formal review because the PR is still in draft....

gschwend:
@johannct  this is a *draft* pull request that I opened as a way to get help from Sam, just to facilitate making comments in the code. Actually, I'd like to withdraw it and to make another one when all the steps described [here]( https://github.com/LSSTDESC/RAIL/issues/4) are completed. I just don't know how to do that. Do you know how can I remove, or give up a request?

johannct:
even then, you do not want the PR to aggregate too much code, else the review will become exceedingly difficult. Whatever the status of the code compared to what is planned, a successful reproduction of the DC1 results should be largely enough for an official PR. If you look at the top of your PR, it states " +47,595 −2", which means that this PR corresponds already to 47595 added lines of code and 2 removed lines of code..... This is a huge ratio, even if I understand well that the code is plain new. So really, I strongly advise to open a PR asap, if possible with only the codebase that allow to reproduce the DC1 results.

johannct:
You can definitely delete a pull request : at the bottom of the page you have a "Close with comment" button. You can also leave it open and start a new one..... with a rebased master or not

gschwend:
Closing the draft pull request. This is a draft pull request that I opened as a way to get help from Sam, just to facilitate making comments in the code. All issues were solved, but the code is not ready to be merged at this stage.

aimalz:
I'm reopening this because #61 needs to merge into the branch for this PR (as soon as the demo stuff is consolidated into the examples directory), and then this one will be ready to merge into the master branch.

rail/evaluation/metrics/cdeloss.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+from rail.evaluation.evaluator import Evaluator
+from rail.evaluation.utils import stat_and_pval
+
+
+default_grid = np.linspace(0., 1., 100)
```
aimalz: Why is this defined here? I'm not used to seeing global variables in a module, and it doesn't seem to be used anyway.

rail/evaluation/metrics/__init__.py
```diff
@@ -0,0 +1,18 @@
+import os
+
```
aimalz: I know we had to do this for estimation (though I hope there's a cleaner fix long-term), but is this necessary for evaluation? Why does it need to be more complicated than, e.g. [the qp init](https://github.com/LSSTDESC/qp/blob/main/qp/__init__.py)?

docs/notebooks/Test_rearranged_evaluation.ipynb
```diff
@@ -0,0 +1,416 @@
...
...3/UNwGNT7oMkNWvaof8vwKYkFyU5C9gG7JtyHySpWVNd3qmqE0l+D/g0sAr4SFUdnGCTIy8RzSDH3IbWxtzaeGFCY07VU5bUJUkrlN/IlaSGGPqS1JAVEfqnO7VDej7Q3f6VJK9cjn6OywDj/c1unF9J8vkkL1+Ofo7ToKfvSPIzSU4meeM0+zcJg4w5yeuS3JvkYJJ/nHYfx22A5/aPJvmbJF/uxvy25ejnuCT5SJJjSe5f5PbxZ1dVzfQ/em8I/zvwIuAs4MvA5lPqXAF8kt73BC4F7lrufk94vK8G1nTbr5/l8Q465r56nwH+Dnjjcvd7Co/zC4AHgAu76+ctd7+nMOZ3A+/pttcC3wTOWu6+jzDm1wKvBO5f5PaxZ9dKmOkPcmqHrcBN1XMn8IIk66bd0TE57Xir6vNV9a3u6p30vg8xywY9fcfbgY8Dx6bZuQkZZMy/AdxWVUcAqmrWxz3ImAt4XpIAz6UX+iem283xqarP0RvDYsaeXSsh9Bc6tcP6IerMiqWO5Wp6M4VZdtoxJ1kP/BrwoSn2a5IGeZxfDKxJ8tkk9yR569R6NxmDjPlPgZ+i96XO+4B3VtWT0+neshh7dk37NAyTMMipHQY6/cOMGHgsSX6RXuj//ER7NHmDjPl9wLuq6mRvEjjzBhnzauBVwBbgWcAXktxZVf826c5NyCBjvgy4F/gl4CeBO5L8U1V9Z8J9Wy5jz66VEPqDnNphJZ3+YaCxJPlp4Ebg9VX1jSn1bVIGGfMccEsX+OcCVyQ5UVV/PZUejt+gz+uvV9X3gO8l+RzwcmBWQ3+QMb8NuL56C96HkzwMvBS4ezpdnLqxZ9dKWN4Z5NQO+4C3du+EXwp8u6qOTrujY3La8Sa5ELgNeMsMz/r6nXbMVXVRVW2sqo3ArcDvznDgw2DP69uBX0iyOsmz6Z2x9tCU+zlOg4z5CL2/bEhyPvAS4KtT7eV0jT27Zn6mX4uc2iHJb3e3f4jepzmuAA4D/0NvtjCTBhzvHwE/BtzQzXxP1AyfoXDAMa8og4y5qg4l+RTwFeBJ4MaqWvCjf7NgwMf5j4GPJrmP3tLHu6pqZk+5nORm4HXAuUnmgeuAZ8DkssvTMEhSQ1bC8o4kaUCGviQ1xNCXpIYY+pLUEENfkhpi6EtSQwx9SWrI/wEAodM9sBLyFAAAAABJRU5ErkJggg==\n",
```
aimalz: I think we've been a bit inconsistent with clearing output in demos vs. leaving it in, so this is mostly a reminder to revisit that later before declaring the Golden Spike done.

aimalz (approved):
At risk of rushing, I'm approving this PR, with some caveats. I've noted a few things that I think should be changed/simplified but ultimately can be dealt with in a subsequent PR if it can't be done trivially now.

rail/evaluation/metrics/cdeloss.py
sschmidt23: You're right, I ran flake8, which usually warns about unused variables, and this isn't flagged by flake8 for some reason.  Removing.

rail/evaluation/metrics/__init__.py
sschmidt23: I think you're right, I forgot that this metrics/__init__ was included in an earlier push and didn't think about why it is present.  We need it in estimation to check for the presence of the necessary base codes for some of the estimators (e.g. DESC_BPZ and flexcode), but I don't think any such check is needed for our evaluators.  It looks like we can just delete this __init__ and everything should still work.

docs/notebooks/Test_rearranged_evaluation.ipynb
sschmidt23: I'll clear the notebook and push as part of this PR.

sschmidt23: Actually, on second look, this was a notebook to illustrate a problem with the spline ensemble rounding off the PIT boundary values, which was remedied when I switched the PIT ensemble storage to quantile rather than spline.  So, I'm just going to remove this notebook from git.

PR #57, opened 2021-03-29
Issue/49/delight

sylvielsstfr:
I think it is time to proceed to do a merge  of delightPZ in master.

The merging is purely functionnal. It means Delight hyper parameters are not optimized for test-DC2 dataset. I will need RAIL metrics for this or doing something else.

Note to integrate Delight with DESC it is necessary to review https://github.com/sylvielsstfr/Delight

A first step is to install the Delight branch desc_rail here  https://github.com/sylvielsstfr/Delight/tree/desc_rail in 
https://github.com/LSSTDESC

Then most of the work of interfacing Delight with RAIL can be found in:

https://github.com/sylvielsstfr/Delight/tree/desc_rail/interfaces/rail

where it is possible to read the README.md file.

Let me know when you are ready for presenting/introducing  how all this stuff works.

Sylvie

### DIFF
--- a/examples/HowToDelightPZ.md
+++ b/examples/HowToDelightPZ.md
@@ -2,46 +2,24 @@
 
 - sylvielsstfr, 
   - creation date : Feb 24th 2021
-  - update : Feb 25th 2021
+  - update : May 8th 2021
 
 ## Delight reference package
 
-To work temporily with the DelightPZ package,
-you have to have installed the Delight
-from github here:
+- You must have a working Delight package installed.
 
-**https://github.com/sylvielsstfr/Delight**
+- Get it from https://github.com/LSSTDESC/Delight
 
-
-and you must select the branch **desc_rail**
-
-Then install Delight according the instructions here
+T- hen install Delight according the instructions here
 
 - **https://delight.readthedocs.io/en/latest/install.html**
 
-- pip install -r requirements.txt
 
-(Notice the coloredlogs has been added)
+Get the Delight data https://github.com/LSSTDESC/Delight/tree/master/data
 
-then
-
-- python setup.py build_ext --inplace
-- python setup.py install --user
+in some path defined in RAIL config file (see DelightPZ.yaml).
 
 
-
-It is important to note that the Delight **setup.py** make that its own data files 
-(FILTERS and SED) are installed somewhere by setuptools.
-
-The user use the setuptools to guess where these data are installed.
-
-The *scripts* in Delight are translated in python modules that can be used by RAIL.
-
-Thus a new module has been created in :
-
-- **Delight/interfaces/rail**
-
-(To avoid too much code inside RAIL)
 
 ## LSSTDESC/RAIL
 
@@ -58,7 +36,7 @@ For the moment, the following files are under developpment in the branch **issue
 
 delightPZ as Delight to generate the Delight 
 configuration file **parametersTest.cfg**
-in a temporary directory (see elightPZ.yaml) .
+in a temporary directory (see DelightPZ.yaml) .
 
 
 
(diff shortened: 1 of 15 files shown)


examples/HowToDelightPZ.md
```diff
@@ -0,0 +1,67 @@
...
+
+- pip install -r requirements.txt
+
+(Notice the coloredlogs has been added)
+
+then
+
+- python setup.py build_ext --inplace
+- python setup.py install --user
+
```
johannct, line 25: in DESC we are advocating to avoid --user as it is extremely fragile. If the packages are not conda installable, one should still work in a conda env and run pip within this env, without --user

examples/HowToDelightPZ.md
```diff
@@ -0,0 +1,67 @@
...
+The user use the setuptools to guess where these data are installed.
+
+The *scripts* in Delight are translated in python modules that can be used by RAIL.
+
+Thus a new module has been created in :
+
+- **Delight/interfaces/rail**
+
+(To avoid too much code inside RAIL)
+
```
johannct, line 45: It is probably not a good idea even if it is a lot of code. And the copying of the utils.py file in Delight, which anyway is going to be deprecated, is a red flag.

tests/data/parametersTest.cfg
```diff
@@ -0,0 +1,74 @@
...
+# - Lines starting with # or ; will be ignored
+# - Multiple values (band names, band orders, confidence levels)
+#   must beb separated by spaces
+# - The input files should contain numbers separated with spaces.
+# - underscores mean unused column
+
+[Bands]
+names: lsst_u lsst_g lsst_r lsst_i lsst_z lsst_y
+directory: /Users/sylvie/MacOSX/GitHub/LSST/Delight/data/FILTERS
+numCoefs: 15
```
johannct, line 14: Problem with personal path here

tests/data/parametersTest.cfg
```diff
@@ -0,0 +1,74 @@
...
+bands_verbose: True
+bands_debug: True
+bands_makeplots: True
+
+[Templates]
+directory: /Users/sylvie/MacOSX/GitHub/LSST/Delight/data/CWW_SEDs
+names: El_B2004a Sbc_B2004a Scd_B2004a SB3_B2004a SB2_B2004a Im_B2004a ssp_25Myr_z008 ssp_5Myr_z008
```
johannct, line 21: personal path

rail/estimation/algos/delightPZ.py
```diff
@@ -0,0 +1,272 @@
...
+from interfaces.rail.makeConfigParam import *  # build the parameter file required by Delight
+from interfaces.rail.processSEDs import processSEDs  # build a redshift -flux grid model
+from interfaces.rail.templateFitting import templateFitting
+from interfaces.rail.simulateWithSEDs import simulateWithSEDs # simulate its own SED in tutorial mode
+from interfaces.rail.delightLearn import delightLearn
+from interfaces.rail.delightApply import delightApply
+from interfaces.rail.convertDESCcat  import convertDESCcat   # convert DESC input file into Delight format
+from interfaces.rail.convertDESCcat  import *   # convert DESC input file into Delight format
+from interfaces.rail.calibrateTemplateMixturePriors import *
+from interfaces.rail.getDelightRedshiftEstimation import *
```
johannct: We probably need to move most of these to RAIL

johannct (changes requested):
@sylvielsstfr ok very first look at it. There is a lot to digest! Would it be possible to add some basic unit tests so that we clear some of the codecov warnings, and we get to see a bit better the basic functionalities?

aimalz:
Are these changes being reflected in [the DESC fork of Delight](https://github.com/LSSTDESC/Delight/pull/3)? I think it would be easier to make the changes there and then have RAIL just install and call from that code, no?

Issue #51, opened 2021-02-26
User-oriented demo with estimation subpackage


aimalz:
The current example code for `rail.estimation.Estimator() `and its subclasses is in a script ([here](https://github.com/LSSTDESC/RAIL/blob/master/examples/main.py)), whereas the example code for rail.creation ([here](https://github.com/LSSTDESC/RAIL/blob/issue/7/creation-baseline/examples/creation/basic-creation-demo.ipynb) and [here](https://github.com/LSSTDESC/RAIL/blob/issue/7/creation-baseline/examples/creation/degradation-demo.ipynb)) and rail.evaluation ([here](https://github.com/LSSTDESC/RAIL/blob/issue/4/evaluation-baseline/rail/evaluation/dc1_check.ipynb) and [here](https://github.com/LSSTDESC/RAIL/blob/issue/4/evaluation-baseline/rail/evaluation/demo.ipynb)) are more interactive Jupyter notebooks. This issue is for migrating the content of the rail.estimation demo script into a notebook, while also test driving the intuitiveness of the config dict/file structure and i/o interface from both the perspectives of someone seeking to use an estimator to obtain some photo-z posteriors and of us as experimenters seeking to benchmark many estimators under multiple sets of input data and configs.

sylvielsstfr:
Delight has a kind of tutorial mode in which it generates its mock fluxes in bands for galaixes at random redshift.

Its usefull to have for debugging at expert level, not for standard user.

Do we want to have this mode to be activated from RAIL ?

johannct:
How would you "activate" it? Can you provide an example of how to call this mode and use it?

sylvielsstfr:
I have a flag in the yaml config file which call this tutorial mode

https://github.com/LSSTDESC/RAIL/blob/issue/49/delight/examples/configs/delightPZ.yaml.

For the moment the python file 
https://github.com/LSSTDESC/RAIL/blob/issue/49/delight/rail/estimation/algos/delightPZ.py

works only in that mode. As soon as I can validate the analysis of Tutorial mode provided by Boris through, I can set this tutorial mode to False and read the true training/evaluation I will start to inject DC2 fluxes inside Delight.

But before I will need to be sure I can use the redshift grid proposed by RAIL inside Delight.

I let you decide at which level of the code development you want to control the code by a pull request.
I am not sure this merge now is usefull because RAIL users would need to install my version of Delight.

[...]

Issue #48, opened 2021-01-29
Modify estimation ingest to remove assumption that we are reading in from a file


sschmidt23:
Current RAIL/estimation is coded such that it assumes we are reading from an hdf5 file.  We should separate the read-in and instead have the data ingest as a separate stage and feed Estimator dictionaries for the test and (optional) training data.

aimalz:
I think the scope of this could be generalized to moving i/o to a standalone module shared across creation, estimation, and evaluation. Then we'd guarantee their mutual consistency no matter how many times we refactor the formats (probably unavoidable given XWG dependencies, etc.), and those refactorings would only have to touch one file. Thoughts?

sschmidt23:
closed by #60

Issue #33, opened 2020-11-30
Switch RAIL/estimation to output qp format rather than hdf5


sschmidt23:
We should switch the outputs from the current hdf5 format to instead spit out qp compatible astropy tables.  RAIL/estimation currently has an iterator to process data in chunks, it looks like the astropy hdf5 tables allows for sequential writes via a file append so I think we can use that for now and maintain iterator functionality.  However, I'm not sure how well this will work with e.g. mpi4py when the chunks may be returned out of order, so we may have to redesign this to something more sophisticated for parallel processing in the near future.

sschmidt23:
In implementing a change where we create qp Ensembles in the native format within RAIL/estimation and then write these to astropy hdf5 tables, I discovered that astropy Table `append` does not actually allow you to append to a single table, instead you can only append to additional 'path' tables within a single file.  This is good and bad: we can write the data in chunks with names where we can recover the original file order even if done out of sequence e.g. when done with multiple processors. But, we would have to do an additional post-processing step to astropy Table `vstack` the chunks into a single file after the code runs to get things into the final format that we want.  I'm now wondering if converting to qp within RAIL/estimation subclasses is worth it, or whether it might be more efficient to just have it as a post processing stage?  Doing it in the subclass and returning a qp Ensemble does force the code writer to make sure that qp functions properly, though we may take a slight performance penalty in the final stage of reading the data back in from file and merging the path chunks.  

Just something to think about, this may change if I find a smarter way to do things.

aimalz:
I think we still want to have the RAIL i/o (well, input only for evaluation and output for both creation and estimation) use qp parameterizations, so maybe this issue is degenerate with proposing hdf5 as an option in LSSTDESC/qp#15? This also ties into the definition of shared data formats for intermediate files (that will be used in ways that are impractical for a database), which @joezuntz has been thinking about for TXPipe as well and which might (soon?) lead to another DESC repository so RAIL doesn't have to handle it at all. Meanwhile, am I correct in understanding that #17 aims to address the aspects of this issue that don't have to do with actual files @johannct? If this issue is indeed degenerate with those two, then perhaps it can be closed?

[...]

PR #60, opened 2021-05-05
Issue/48/fileIO PR that addresses issue #33, issue #48, and issue #51 

sschmidt23:
This PR covers several issues, as pointed out in the title:

issue #33: adding qp format to outputs.  This PR adds an `output_format` keyword to the base.yaml, that, if set to "qp" writes out PDFs in qp format rather than an hdf file (you can still write to hdf5 by setting `output_format` to "old" or "hdf5" or anything other than "qp").

issue #48: this PR modifies the assumption that data is read in from a file for the training data.  However, on talking with @aimalz he mentioned that he did think that the test data should still be assumed to be read in from a file as specified by `testfile` in the base_config yaml/dictionary.  It moves the former functions in rail/estimation/utils.py to rail/fileIO.py, as they may be of more general use to other rail packages for reading/writing hdf5/qp/pandas etc... files.

Aspects of these first two modifications can be easily tested by running the quick demo script in the examples directory with e.g.
`python main.py configs/FZBoost.yaml`

issue #51: This PR contains a RAIL/estimation user demo notebook in doc/notebooks/RAIL_estimation_demo.ipynb.  There are a few placeholder bits, for example a `qp-append` function that should actually live in `qp` and some plots of mode and "summed N(z)" that are nice visual demonstrations/sanity checks, but should not be used for science that I'm not sure of the best way to present.  

To address a comment from Bryce about default parameter values, I added a default_param dictionary for each of the existing codes as well as a check that will run with the defaults if no dictionary is supplied, and also check to see if any parameters are missing from the config_dict and replace with default if one or more is missing.  There is also a description_dict that describes the parameters that is printed along with the current values when a specific subclass is called.


doc/notebooks/RAIL_estimation_demo.ipynb
```diff
@@ -0,0 +1,644 @@
+{
```
johannct: I experienced an issue in running the notebook : the setup.py installed scikit-learn 0.23.2 somehow, and the notebook fails with this version. I had to conda install scikit-learn to install the more recent 0.24.2 version and the notebook ran successfully with it. So we may need to enforce sklearn >0.24 somehow.... I am not exactly sure why I did not get the latest sklearn though......

doc/notebooks/RAIL_estimation_demo.ipynb
```diff
@@ -0,0 +1,644 @@
...
+   "metadata": {},
+   "source": [
+    "Yes, under a second.  So, if you are running an algorithm with a burdensome training requirement, saving a trained copy of the model for later repeated use can be a real time saver."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Add a temp qp append function!"
```
johannct: ok let's close this PR and then we will open an issue about the qp_append in qp

examples/base.yaml
```diff
@@ -4,4 +4,5 @@ base_config:
   hdf5_groupname: photometry
   chunk_size: 1000
   configpath: ./configs
-  outpath: ./results
\ No newline at end of file
+  outpath: ./results
+  output_format: old
```
johannct, line 8: This output_format raises a more general issue in my mind : qp is not a util storage library but a pdf representation library. It happens to be able to transparently persist any of its pdf representation. But it may very well be modified so that the old way to store the rail results are replicated in qp, so that this old/qp distinction is not relevant anymore. Testing storage through qp is certainly important (and I think should be a test case), but we need to show in the notebook that we are able to choose the representation we want to persist, be it interp, sparse, or quantiles. 
We can certainly open a separate issue and deal with it once this PR is merged : it would be a global issue about proper interfacing to the qp library, now that the latter is getting more mature.

tests/test_filIO.py
```diff
@@ -0,0 +1,68 @@
+import pytest
```
johannct: test_filIO and note test_fileIO? A detail, admittedly :)

johannct (approved):
OK LGTM. There are comments that need separate handling in future issues, that I would appreciate comments here before merging/deleting, if any. The one issue I would like to understand better is the sklearn version issue.....

Oh one final question : you put the notebook inside a `doc` directory? Is that meant to avoid a collision with the `docs` directory created by running the doc building, or is it just an oversight? Unless it really is a problem, we should have only one `doc` or `docs` directory.

doc/notebooks/RAIL_estimation_demo.ipynb
sschmidt23: Ah, I have v0.24.1 on my home machine where I did testing, so I didn't see any problems.  I'll add a >=0.24.1 to the setup.py

examples/base.yaml
sschmidt23: Right now I have it set up so that you do not choose the representation that you persist, but rather it writes out the particular qp parameterization that is encoded in each subclass's `estimate` function.  You could, of course, modify the `estimate` function to spit out a different parameterization by doing a transform, but I do not have the output parameterization as a variable or anything.  We could write a utils function to do such a transform before saving, but it would requite a little bit of rearranging of things.  I *think* what Alex wanted was to write out the output in the code's "native" qp parameterization, which is what I went for here.

doc/notebooks/RAIL_estimation_demo.ipynb
sschmidt23: I'll bring this up with Eric, there are several qp-related things that need updating with evaluation as well, so I think a qp PR will be coming in the near future.

examples/base.yaml
johannct: ok but no estimator code tries to be smart in terms of pdf storage..... So eventually we will need to do something.

johannct: Can we have a table somewhere of how the DC1 (for instance) codes output the pdf?

sschmidt23: For DC1 we *forced* all codes to an interp parameterization (this was Alex Abate's idea to simplify things, as this was before qp existed, and was, I think, one of the motivations for starting qp).  So, the table for DC1 is easy: every entry is interp.

tests/test_filIO.py
sschmidt23: Fixed

examples/base.yaml
johannct: ok so I am still not sure where we want to go here.... How do we intend to persist the pdf uniformly across all the executed estimators? We can't have one distinct representation per estimator stored, can we?..... so either we continue forcing everybody to return interp, and then we store interp for now (costly in terms of storage, and as you said qp has been started to prevent this eventually), or we call qp to transform into a single storage representation every estimators' outputs.

doc/notebooks/RAIL_estimation_demo.ipynb
sschmidt23: Actually, I forgot that I already created a qp issues for this: https://github.com/LSSTDESC/qp/issues/24

doc/notebooks/RAIL_estimation_demo.ipynb
sschmidt23: Hmm, the GitHub action can't find a pip installable sklearn >=0.24, so I am going back to removing this.  What error did you get on 0.23.2?  All of the tests run in the GitHub action with no problems, though this of course does not run the notebook.  There is a non-convergence error that pops up intermittently with the neural net in simpleNN, but this occurs in *all* versions of sklearn.  It's more the simple architecture and only 10k training galaxies does not always converge in the default number of iterations.  I think that's more on the data than on the algorithm.

sschmidt23:
For the comment above about doc/notebooks vs docs/notebooks, that was a typo on my part.  I have moved the estimation demo notebook to docs/notebooks.  

Other than the potential concern about sklearn >=0.24 I think all issues are taken care of.  I'll wait for one last comment on that from you before merging this in.

examples/base.yaml
sschmidt23: Wait, but if we set the format to `qp` then the estimator will write out to qp in whatever format is specified in the subclass.  For example, right now `FZBoost` writes out to interp, `simpleNN` writes out to a scipy.stats.norm (Gaussian), `trainZ` writes out to a interp, etc...  qp can auto-detect the type of the data when you do a qp.read(infile) based on the metadata stored in the associated infile_meta file.  

I don't think that we have to worry about persisting the pdf uniformly, that's the whole idea of qp.  It should use whatever format is supplied to it, no need to be the same across codes.  @aimalz could maybe chime in here with any comments.

sschmidt23: Or, I should say more exactly, `estimate` will return a qp-formatted object, and there are functions to write out the data (in chunks or all at once) in rail/fileIO.py that can be used as they are, for example, in examples/main.py

johannct: So yes you are right about qp functionalities here. It can be done this way. Is it what we want? Ultimately we will need to be economical in storage space, and no estimator I know of natively outputs pdf in quantile or sparse representation. Fine if we do not care for now, but imagine a subsequent stage of hybrid estimation where say we use template fitting in some part of color space and ML in some other, and it happens that each outputs pdf in different format? My only point is that it is hard for me to imagine that we are not cornering ourselves with just letting every estimators persist in their native format. Of course we can consider that it will always be possible to reopen all the files and run qp over them to persist in a unique way ultimately....

doc/notebooks/RAIL_estimation_demo.ipynb
johannct: I had a serious API issue, not a fitting fluke. Let's leave that aside for now : we'll deal with it if it bites us again.

johannct:
yes you can go ahead and merge : I put two final comments above, but nothing barring merging. For the record I installed scikit-learn with conda, and we should prefer conda installation rather than pip within DESC, when possible.

examples/base.yaml
aimalz: Re: what we want, we can introduce a mechanism in the estimation process, like a parameter in the shared-across-estimators config file, to impose limits on the number of stored parameters per galaxy. However, it's definitely outside the scope of this PR. At this stage, we're also assuming unlimited computing resources, which of course isn't viable either -- both are things we'll have to systematically test. I'll make a new issue for this, because it does touch on a bigger question of how to provide estimators with storage-related parameters when their native formats may differ considerably.

sschmidt23: I think it would be pretty trivial to add a conversion step in whatever script that we use to generate any catalog-level things to convert a native distribution to whatever "final" storage format we want.  It should literally be a one-liner with an `ensemble.convert_to` call.  I like the idea of preserving the "native" format output by each code as long as possible to prevent any losses due to potential transforms.

aimalz: Yeah, at the moment, given the scale of tests planned with just the Golden Spike v1 code, it's more important to store the native format and prevent information loss due to conversion. For the sake of reproducibility/follow-up of the original qp paper, we should also be prepared to have the shared-across-algorithms config file specify the number of parameters that can be stored per posterior though, regardless of the parameterization. I just thought this discussion might benefit from explicitly recognizing parameterization and number of parameters as distinct contributors defining the otherwise overly broad concept of "storage format."

PR #61, opened 2021-05-25
Issue/4/evaluation rearrangement

aimalz:
We took a stab at rearranging the evaluation subpackage for the _Golden Spike_ v1 release. The initial goal of this work was to isolate the functionality that we hope to ultimately call through qp, hopefully making that future transition a little easier. More of the changes, however, ended up being related to enforcing some of the API conventions established in the estimation subpackage, specifically those affecting contributors adding new metrics and those affecting how we as experimenters will use it for the baseline tests of the _Golden Spike_. We should converge on these before getting too far into unit testing.

We got as far as the PIT and metrics thereof (modulo a snag with [unexpected behavior of qp](https://github.com/LSSTDESC/qp/issues/33)) and are working on a demo notebook as well as a similar rearrangement for the CDE Loss -- changes will indeed have to be propagated to the plotting functionality as well, but that can be done back in the primary evaluation branch, along with the unit tests, once this patch is merged in.


aimalz:
@Gschwend I left a reminder to myself to post some to-do items to this issue following our discussion yesterday, but unfortunately the only ones I wrote down then were easy enough that I already did them, plus one about moving the DC1-specific code, demo notebook, and demo `rail.evaluation.main.py` script from `RAIL/rail/evaluation/plot_utils.py` to the `RAIL/examples` directory. I think the plotting code might also belong there? Sorry for not posting that right away. I think the other main things were to propagate the new structure through the demos and plotting code and to write the unit tests.

aimalz:
Re: consolidating the location of example code, @sschmidt23 since you wrote the estimation demo of #51 and the demo of the new evaluation structure, would you mind pulling the post-#60 changes from master, moving those two demos to (appropriate subdirectories of) the examples directory, and checking that they still work in the new location? That might help with giving @Gschwend an obvious place to move the evaluation plotting code and DC1-specific demo material next.

sschmidt23:
So are you advocating for a move of doc/notebooks to examples/notebooks?  File organization may be a larger issue than just the examples, as I just pushed a [BPZ](https://github.com/LSSTDESC/RAIL/tree/issue/64%2FBPZ) branch that has some extra files required by BPZ, plus the SEDs and Filter curves.  So, I was wondering if we should open a separate, dedicated, issue to cover the structure of where files "live" and such.  For example, in [issue/64/BPZ](https://github.com/LSSTDESC/RAIL/tree/issue/64%2FBPZ), right now there is a folder called "TMPBPZ" in examples that is an egregiously inappropriate name on purpose to remind me to move it to a better location.

sschmidt23:
Ah, I forgot that creation already has its own sub-folder under examples in the main branch, so the logical thing is to create estimation and evaluation folders and stick everything in there.  I still think that an overall file cleanup is better suited to its own issue than being part of this PR.  But, I would say for now, I would create an `examples/evaluation` folder and stick the demo notebook there in this branch, which will minimize conflicts with the file reorganization issue that I'm about to create.

johannct:
CLMM has an examples directory, and a demo directory. I would imagine that we can subdivide the usage examples under examples/evaluation, examples/creation and examples/estimation, but the demos, or pipelines, or end to end tutorial of RAIL should not be subdivided unside the demos subdirectory. Now of course, the divide between demos and examples can be blurry.....

gschwend:
@aimalz  this branch is ready to merge with the evaluation baseline. I already moved the main.py to the examples directory and tested it there.

gschwend:
I'm not sure if I'm the one supposed to press the merge button. Just in case, I'll wait for people to take a look at the code changes added by me this week.

PR #67, opened 2021-06-04
Issue/66/ex org

sschmidt23:
This PR reorganizes the examples directory, and merges in the docs/notebooks directory.  

I just tested that the estimation notebook works when run on cori, as does the command-line `examples/estimation/main.py` when run with the various `configs/` files.

Is this structure what we were aiming for?


johannct (approved):
LGTM. You might want to wait for the doc PR to be merged so that you can rebase and get the examples notebooks properly rendered.

Issue #64, opened 2021-06-01
Add BPZ estimator to RAIL/estimation


sschmidt23:
We don't need full functionality producing things like the "best fit type" and "best type without prior", so I'll call it `bpz_lite` to distinguish from "full" BPZ (though I just added the python3 version of BPZ to LSSTDESC as `DESC_BPZ` so the "full" BPZ is available for running from the command line from there anyway).

sschmidt23:
closed by #70

PR #70, opened 2021-06-17
Issue/64/bpz addresses  issue #64

sschmidt23:
This PR adds `bpz_lite`, a simplified subset of BPZ functionality, to RAIL.  It uses some functions from a new `DESC_BPZ` (https://github.com/LSSTDESC/DESC_BPZ) package that I added, and which is now included in the `setup.py`.  A lot of the code is taken from BPZPipe, but I also added back in some features that were missing, e.g. the ability to compute AB files if they are not present.  Adam Broussard also identified a bug in Madau reddening in (pretty much all version of) BPZ, I implemented a change to `DESC_BPZ` that fixes that problem.  Whoever reviews this PR may also want to look at the `DESC_BPZ` package, it's basically just a slight rearrangement of the python3 version of BPZ that we have been using but reorganized to look more like a python package.

Major features: 
- I added a new folder: `rail/estimation/data` to the directory structure that contains subdirectories `FILTER` containing the Filter transmission files, and `SED` containing the SED template files.  Template-based codes should be able to read from this location, that way we can specify *exactly* which SEDs and FILTERs are being assumed when running the code (as opposed to codes that grab data from an outside location that is hardcoded somewhere else).  There is also an `AB` directory that is needed by `bpz_lite` to store some temporary files.  
- A few new config files are needed by BPZ, which are now in rail/estimation/configs/test_bpz.columns(specifies the filters and photometry to use) and rail/estimation/data/SED/CWWSB.list (specifies the SEDs to use)
- For now I only copied the traditional "CWWSB" SEDs to the directory, we can of course add additional SED templates in the future, but this keeps things lightweight.
- I added the LSST filter curves that were assumed when constructing cosmoDC2 to rail/estimation/FILTER.  There are other slightly different versions of LSST filters floating around, so we'll need to be careful to keep track of the exact filter set that is assumed for different data.


examples/estimation/base.yaml
```diff
@@ -5,4 +5,4 @@ base_config:
   chunk_size: 1000
   configpath: ./configs
   outpath: ./results
-  output_format: old
+  output_format: qp
```
johannct: It is probably fine and the easier path, but I still ask so that it is documented with the PR : is it necessary or even a good thing to throw qp integration in with this PR?

examples/estimation/configs/BPZ_lite.yaml
```diff
@@ -0,0 +1,17 @@
+run_params:
```
johannct, line 1: see #71

examples/estimation/configs/simpleNN.yaml
```diff
@@ -1,7 +1,7 @@
 run_params:
   class_name: simpleNN
   run_name: test_simpleNN
-  width: 0.065
+  width: 0.025
```
johannct: is that change justified? it is currently one of the 3 conflicts blocking the merge

examples/estimation/configs/test_bpz.columns
```diff
@@ -0,0 +1,11 @@
+#Filter        columns  cal(AB or Vega,optional)  zp_error  zp_offset
```
johannct, line 4: add a comment header to shortly explain what this file is used for

rail/estimation/algos/bpz_lite.py
```diff
@@ -0,0 +1,376 @@
...
+                print(f"name for model not found, will save to {defModel}")
+                self.inform_options['modelfile'] = defModel
+        datapath = inputs['data_path']
+        if datapath is None or datapath == "None":
+            railpath = os.path.dirname(rail.__file__)
+            tmpdatapath = os.path.join(railpath, "estimation/data")
+            os.environ["BPZDATAPATH"] = tmpdatapath
+            self.data_path = tmpdatapath
+        else:  #pragma: no cover
+            self.data_path = datapath
```
johannct, line 161: check path validity

rail/estimation/algos/bpz_lite.py
```diff
@@ -0,0 +1,376 @@
...
+        # make a list of all available AB files in the AB directory
+        ab_file_list = glob.glob(os.path.join(data_path, "AB")+"/*.AB")
+        ab_file_db = [os.path.split(x)[-1] for x in ab_file_list]
+
+        for i, s in enumerate(spectra):
+            for j, f in enumerate(filters):
+                model = f"{s}.{f}.AB"
+                if model not in ab_file_db:
+                    self.make_new_ab_file(s, f)
+                model_path = os.path.join(data_path, "AB", model)
```
johannct, line 200: define AB dir at like 188 or even immediately after data_path definition, so that if we decide to move these dirs around the hardcoded changes in the file will be easier to track down.

rail/estimation/algos/bpz_lite.py
```diff
@@ -0,0 +1,376 @@
...
+                zo, f_mod_0 = get_data(model_path, (0, 1))
+                flux_templates[:, i, j] = match_resol(zo, f_mod_0, z)
+
+        return flux_templates
+
+    def make_new_ab_file(self, spectrum, filter):
+        from desc_bpz.bpz_tools_py3 import ABflux
```
johannct, line 207: I do not mean to review DESC_BPZ, but is it still necessary to have the `_py3` here?

rail/estimation/algos/bpz_lite.py
```diff
@@ -0,0 +1,376 @@
+"""
+Port of *some* parts of BPZ, not the entire codebase.
+Much of the code is directly ported from BPZ, written
+by Txitxo Benitez and Dan Coe (Benitez 2000), which
+was modified by Will Hartley and Sam Schmidt to make
+it python3 compatible.  It was then modified to work
+with TXPipe and ceci by Joe Zuntz and Sam Schmidt
+for BPZPipe.  This version for RAIL removes a few
+features and concentrates on just predicting the PDF.
+
```
johannct, line 10: This acknowledgement should be in DESC_BPZ, together with any licensing information.

rail/estimation/algos/bpz_lite.py
```diff
@@ -0,0 +1,376 @@
...
+        # The likelihood and prior...
+        pczt = p_c_z_t(flux, flux_err, flux_templates)
+        L = pczt.likelihood
+
+        # old prior code returns NoneType for prior if "flat" or "none"
+        # just hard code the no prior case for now for backward compatibility
+        if prior_file == 'flat' or prior_file == 'none':
+            P = np.ones(L.shape)
+        else:
+            P = prior(z, mag_0, prior_file, nt, ninterp=0)  # hardcode interp 0
```
johannct, line 289: I would write a `set_prior` function, just in case this step becomes more involved with time. I could imagine a user literally providing his/her own python module after all. And this will probably occur anyway when `inform` is implemented.

rail/estimation/algos/bpz_lite.py
```diff
@@ -0,0 +1,376 @@
...
+        flux_err = np.sqrt(flux_err**2 + add_err)
+
+        # Upate the input dictionary with new things we have calculated
+        data['flux'] = flux
+        data['flux_err'] = flux_err
+        data['mags'] = mags
+        return data
+
+    def estimate_pdf(self, flux_templates, kernel, flux, flux_err, mag_0, z):
+        from desc_bpz.bpz_tools_py3 import p_c_z_t, prior
```
johannct, line 273: is it really a good thing to import here rather than at the top of the file? Is it even PEP8 compliant? I never remember.....

rail/estimation/algos/bpz_lite.py
```diff
@@ -0,0 +1,376 @@
...
+        # Loop over all ng galaxies!
+        for i in range(ng):
+            mag_0 = test_data['mags'][i, m_0_col]
+            flux = test_data['flux'][i]
+            flux_err = test_data['flux_err'][i]
+            pdfs[i], zmode[i] = self.estimate_pdf(self.flux_templates,
+                                                  kernel, flux,
+                                                  flux_err, mag_0,
+                                                  self.zgrid)
+
```
johannct, line 376: ok the for loop obviously will go away, and its proper multi-core/node distribution should be taken care of by the base class logically. For the time being, I would recommend bringing `self.flux_template` to local scope only once (I think calling `self.` repeatedly is an overhead), and extracting the test_data columns in one shot, and then define the loop as `for i,m0,f,fe in enumerate(zip(m0_col,f_col,fe_col):` or something like that

setup.py
```diff
@@ -32,6 +36,7 @@
                       'pyarrow',
                       'tables',
                       'astropy',
+                      'DESC_BPZ @ git+https://github.com/LSSTDESC/DESC_BPZ',
```
johannct, line 39: any chance you can take the time to turn it into a pip installable package?

johannct (changes requested):
- I think that `base.yaml` is now broken :  the paths to trainfile and testfile need to go two level up : `../../`
- irrespective of what the structure will eventually be and how we will deal with provenance, the new dirs AB, SED, FILTERS need each a README.md file explaining where the various files come from and what there nominal usage is. Else in the interim this may still drive us crazy.
- 3 files have a conflict to solve.
Thanks a lot for all this! Let's clean the comments fast so that we can approve.

examples/estimation/configs/BPZ_lite.yaml
```diff
@@ -0,0 +1,17 @@
...
+  class_name: BPZ_lite
+  run_name: BPZ_lite
+  zmin: 0.0
+  zmax: 3.0
+  dz: 0.01
+  columns_file: ./TMPBPZ/test.columns
```
johannct: this seems to be needed but is not provided in the current branch, so I cannot run `python main.py configs/BPZ_lite.yaml`

examples/estimation/base.yaml
sschmidt23: I flipped this default while doing some testing and forgot to change it back, I'll revert it back to `old` now.

examples/estimation/configs/simpleNN.yaml
sschmidt23: I was playing with some tests and didn't mean to commit this change, reverting.

examples/estimation/configs/test_bpz.columns
sschmidt23: comment added

rail/estimation/algos/bpz_lite.py
sschmidt23: It's a remnant from when Will Hartley was working on the update, I think.  It does distinguish it from the py2 version that is floating around.  I'm not bothered by it, but I also wouldn't mind removing the _py3 from filenames.

rail/estimation/algos/bpz_lite.py
sschmidt23: It actually belongs in both, as some of the BPZ code functionality is copied in bpz_lite as well.  I do need to add more specific acknowledgements to DESC_BPZ after I email Dan Coe to ask a couple of questions.

rail/estimation/algos/bpz_lite.py
sschmidt23: BPZ already reads in a prior from a file specified in the config file (in this instance it is `hdfn_gen`, which reads the python function name `prior` from `prior_hdfn_gen.py`.  That is, the code defines prior as an import from a specific file name, so this already functions as if it was `set_prior` rather than just `prior`

rail/estimation/algos/bpz_lite.py
sschmidt23: This is left over from some code that Joe Zuntz wrote, but yes, I'll move all imports to the top of the file.

setup.py
sschmidt23: I asked Alex about this, I think the consensus was that not all installable packages need to be on PyPI, this setup pulling from GitHub seems to work, so we may leave it in that form rather than insisting on all packages being on PyPI.

examples/estimation/configs/BPZ_lite.yaml
sschmidt23: Ah, this was renamed in an internal reorganization that I did and I missed it.  Changing the columns file to point to examples/estimation/configs/test_bpz.columns

rail/estimation/algos/bpz_lite.py
sschmidt23: Ah, trying to change this reminded me why the imports are structured like this: bpz_tools_py3 requires that an environment variable be set, which happens in the __init__.  If the imports to bpz_tools_py3 are at the top of the file, an exception occurs because the variable has not yet been set.  

So, I reverted the changes and moved the imports back to within the functions in bpz_lite.

rail/estimation/algos/bpz_lite.py
sschmidt23: The for loop is only looping over the subset passed to the code in each chunk, e.g. 1000 galaxies out of ~21,000 in the example test data, so I don't think it will go away.  For flux_template, I don't think that the speed loss is fairly minor, and not worth making a copy locally, though we can try it and see.  For the other variables, I find this form much more readable, and I think it involves less copying of data, since we're only grabbing one row of the test_data, rather than copying the whole thing over and enumerating those.

rail/estimation/algos/bpz_lite.py
johannct: as you like. Maybe an issue to open for the next round of improvement?

rail/estimation/algos/bpz_lite.py
johannct: oh then the acknowledgement and licensing needs to be a bit more formal, then.

rail/estimation/algos/bpz_lite.py
johannct: hmm I have a hard time following the code here, especially in the context of an implementation of `inform`.....

rail/estimation/algos/bpz_lite.py
johannct: ok..... sounds like hiding a problem though...

rail/estimation/algos/bpz_lite.py
johannct: ti should be a view, not a copy. But I am no numpy memory contiguity expert either. I just learned the hard way to work column wise as much as possible, and loop over rows only when strictly necessary. As a matter of fact the called function could take arrays as argument, rather than floats. This would move the for loop closer to actual computations

setup.py
johannct: DESC CO may insist.... Ultimately all packages that will be deployed and integrated by DESC CO may need to be at least pip installable, and CO may do the extra work to turn them intoconda installable. Working the way above is a no go because at some point the version of the package will need to be hardcoded in there, which is not practical at all when we will need to gather all the version of all the package of a given DESC software release.

rail/estimation/algos/bpz_lite.py
sschmidt23: The original BPZ code has no license file.  I'm about to email Dan Coe about the Madau bug fix in DESC_BPZ, I'll also ask if it's ok that we have messed with the code, and also about posting to PyPI.  Again, this being a modification of a public code with no official license is an uncomfortable area of space to be in.  I think there may be similar issues with other codes, e.g. Delight with similar issues (though I don't know what license info Delight has).

rail/estimation/algos/bpz_lite.py
sschmidt23: This is the way the original code was set up, again, the goal here was to do a minimal port of BPZ, and this is how BPZ is written.  For inform, I think it will entail running a custom fit to find the ~14 parameters, and then a special prior function that somehow reads in those parameters.  I haven't worked out exactly how I wanted to do that yet, which is why `inform` remains unimplemented.

rail/estimation/algos/bpz_lite.py
johannct: Yes let's try to clean things up. The advantage with Delight is that Boris is in DESC.

rail/estimation/algos/bpz_lite.py
sschmidt23: Yes, I agree, though the way the original code was written this is how things work.  There's a similar issue with the prior, as the name of the .py file that it imports `prior` from is also in the config file.

rail/estimation/algos/bpz_lite.py
sschmidt23: added a check

rail/estimation/algos/bpz_lite.py
sschmidt23: Ok, I pulled out the two `self.` references in the loop and also changed this to a list comprehension rather than a for loop.

setup.py
sschmidt23: Ok, but that issue is beyond this PR, we can create a separate issue somewhere, as this same issue is also true for qp as at least one other RAIL dependency.

rail/estimation/algos/bpz_lite.py
sschmidt23: I'm waiting on a reply from Dan Coe, but again, this is an issue for DESC_BPZ rather than bpz_lite, so is there anything remaining to be done here?

examples/estimation/configs/BPZ_lite.yaml
sschmidt23: Should we deal with this separate from this PR, as it applies to every yaml file in configs?

rail/estimation/algos/bpz_lite.py
sschmidt23: Any combo of the loop looked uglier and less comprehensible to me than the original for loop, and the original loop is also closer to the original BPZ code, so I switched it back.  I did store local copies of the two self variables that are called in the loop, which should save a small amount of time.

setup.py
johannct: yes

johannct (approved):

rail/estimation/algos/bpz_lite.py
sschmidt23: I'll create an issue on DESC_BPZ and mention the renaming

sschmidt23:
I heard back from Dan Coe via email, he wasn't sure of which License to use and CC'd Noel Benitez on the email (though Benitez has apparently left the field).  I'm not sure how long it will take to hear back on the License issue, which is really an issue for DESC_BPZ not RAIL, so I'm going to resolve the conflict and try to merge this in now, and we can worry about the incomplete readthedocs later.

PR #79, opened 2021-07-14
Let users of the NN and flexzboost select which bands to use

joezuntz:
This makes it possible to use a subset of bands with two of the estimators.

### DIFF
--- a/examples/estimation/configs/FZBoost.yaml
+++ b/examples/estimation/configs/FZBoost.yaml
@@ -4,6 +4,7 @@ run_params:
   zmin: 0.0
   zmax: 3.0
   nzbins: 301
+  bands: ugrizy
   trainfrac: 0.75
   bumpmin: 0.02
   bumpmax: 0.35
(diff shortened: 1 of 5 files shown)


sschmidt23:
This looks fine, I added `bands` as defaults in the two examples as well just for completeness.  At some point soon I need to add in a more flexible naming convention for bands, which are currently hardcoded, but that should not conflict with this functionality.

I'll also note that for BPZ_lite, you can control which bands are used via the `columns` file, e.g. in `configs/test_bpz.columns` file you can comment out lines beginning with particular filters to stop DESC_BPZ from using them in the analysis.

But yes, go ahead and merge this in.

johannct:
Is there a unit test to check the syntax of `bands`? Is it "gri" or "g,r,i" or what?

sschmidt23 (approved):

joezuntz:
I have added a test for what @johannct mentioned above - you now get an error message for non-letter bands.  And there is a unit test to make sure these are correctly raised.

joezuntz:
Thanks all. Malz says I can merge :-)

PR #81, opened 2021-07-27
Fixed the implementation of the `LineConfusion` degrader.

### LINKED ISSUE #80: Fix `LineConfusion` Degrader
I realized that the implementation of the `LineConfusion` degrader isn't quite correct. I will fix that now.

jfcrenshaw:
The original implementation of the `LineConfusion` degrader had a mistake.

Suppose we observe a spectral line at lambda_obs. This observed line is due to redshifting of the spectral line at lambda_true in the rest frame. However, if we misidentify the line, we will think that the line is actually at lambda_wrong in the rest frame.

We can calculate the erroneous redshift, z_wrong, as a function of the true redshift, z_true, and the wavelengths of the true and wrong lines:
image

The original implementation had the ratio lambda_true / lambda_wrong inverted.

I fixed this mistake, then reran the degrader example notebook to propagate the changes there.


sschmidt23 (approved):
Looks good, and everything works.  Go ahead and merge and delete the old branch.

Issue #68, opened 2021-06-07
bring requirements up to date in master


johannct:
fileIO.py is now in master, but it depends on astropy, which did not make its way to the requirements file, at least not explicitly. More generally, we may want to also autogenerate the docs/requirements.txt file, so that the common part of the two files is synchronized.

eacharles:
overtaken by events

PR #83, opened 2021-07-30
Added install instructions

### LINKED ISSUE #82: Add installation instructions to the README and remove the environment yaml
Noticed this when dependabot suggested we update the tensorflow dependency listed there - `Creation` doesn't use tensorflow anymore!

jfcrenshaw:
I also deleted the outdated `environment.yml`.


jfcrenshaw:
I expanded the scope of this pull request a little bit... Usually I try not to do that, but it was on the topic of installation.

I restructured `setup.py` so the required and extra packages are all set at the top. 
I streamlined this so that the list of packages for the full installation is automatically pulled from the lists of packages for creation, estimation, and evaluation.
I also made separate declarations for the base estimation packages, and the packages required for each estimation code.
Now installing `rail[estimation]` will download all the base estimation packages + all the estimators, but installing e.g. `rail[bpz]` will install all the base estimation packages + the packages required for BPZ (and not the other estimators).

I also moved the version number to a `version.py` module, so that the version number can be controlled from one place, and is accessible from within the package as `rail.__version__`.

rail/version.py
```diff
@@ -0,0 +1,5 @@
+# Store the version number here so that
+# 1. We can control the version number in one location
+# 2. It is accessible from the package (see __init__.py)
+# 3. We can access it from setup.py without loading rail.
+__version__ = "0.1.dev0"
```
aimalz, line 5: I guess this can be changed to `v0.9-beta` (and I'll take care of revising the tag).

setup.py
```diff
@@ -1,15 +1,68 @@
 from setuptools import setup, find_namespace_packages
 
-packages = find_namespace_packages()
+# basic dependencies for all RAIL modules
```
aimalz, line 3: Just to check, @johannct , does this effectively address #68?

setup.py
```diff
@@ -1,15 +1,68 @@
...
+
+# dependencies for the Creation module
+creation_extras = ["pzflow"]
+
+# dependencies required for all estimators in the Estimation module
+estimation_extras = [
+    "matplotlib",
+    "pyarrow",
+    "pyyaml",
+    "scipy>=1.5.2",
```
aimalz, line 20: Thanks, almost forgot about this!

aimalz (approved):
Whoops, sorry, I mistakenly thought this had already been merged in when tagging yesterday's commit after #54!

PR #84, opened 2021-08-06
Re-set BPZDATAPATH when unpickling BPZ_lite class

joezuntz:
Currently if you pickle a BPZ_lite class and then unpickle it in a new process (as wo do in TXPipe, since we separate the train and estimate phases) then the BPZDATAPATH environment variable is no longer set and there is an import failure later when importing the desc_bpz library.

This adds a `__setstate__` special method to BPZ_lite that restores it. This is automatically called when you unpickle a class.

### DIFF
--- a/rail/estimation/algos/bpz_lite.py
+++ b/rail/estimation/algos/bpz_lite.py
@@ -381,3 +381,9 @@ class BPZ_lite(Estimator):
         else:
             pz_dict = {'zmode': zmode, 'pz_pdf': pdfs}
             return pz_dict
+
+
+    def __setstate__(self, state):
+        # Set BPZDATAPATH when unpickling so state is restored
+        self.__dict__.update(state)
+        os.environ["BPZDATAPATH"] = self.data_path


sschmidt23:
Is this really what you want?  Won't this reset BPZDATAPATH to the value that was set in the pickled version of the class used to generate the pickled object?  BPZDATAPATH is a (not great, legacy from how BPZ was set up 20 years ago) way to point to the SED, FILTER, and AB directories on the local machine running the code, it should not actually be associated with the trained data.  So, if you created your pickle file on one machine then moved to another with a different location for the SED/FILTER files then I think this would break.

Having the path set as an environment variable is a bit awkward, I think the real fix might be getting rid of this env var dependence in DESC_BPZ and just passing the path in, which would likely solve this problem.

joezuntz:
Hi Sam - that's a very good point, thanks, I hadn't considered moving one of these pickles between machines or users, which is clearly needed! 

I've made a PR (actually two, one adding CI) to desc_bpz instead.

PR #86, opened 2021-08-12
Tweaks to BPZ to match behaviour of script

joezuntz:
This does two things to the BPZ RAIL estimator:
- converts mag = -99 values to being flux=0 and flux_err = 1e108. This matches the bpz.py script
- Changes the clipping in mag_err. The current upper limit allows underflow of flux_err which ends up yielding flux_err = 0 and consequent NAN likelihoods.


sschmidt23 (approved):
Looks good, thanks for catching the missing "non observed" in bpz_lite.  Merging into main

PR #92, opened 2021-09-13
add first pass end-to-end script

### LINKED ISSUE #85: Add a script that does basic "golden spike" end to end analysis
That is, chain together creation, estimation, and evaluation to demo the full pipeline

sschmidt23:
First stab at a golden spike script.  There are several kludge-y bits with the fileIO so I will mark this as a draft PR for now.  I also copied a few of the functions from Julia's examples scripts for plotting and point estimates into `eval_utils.py`, we should discuss whether these should live somewhere more permanent (mainly for the point statistics).  

There is also some placeholder stuff in `spike_utils.py` where I use PhotoZDC1 to add errors to the truth data, if @jfcrenshaw has an updated flow that incorporates magnitude uncertainties and non-detections (or if we switch to luptitudes or some other system), then this will need to be replaced.


sschmidt23:
Oh, and I forgot to run flake8, so there's likely some linting and such to still take care of.

examples/goldenspike/eval_utils.py
```diff
@@ -0,0 +1,271 @@
...
+    lower_h = 3.0 - 3.0*(sigma)
+    plt.plot([0.,3.],[upper_l,upper_h],c='r',lw=1,linestyle='--')
+    plt.plot([0.,3.],[lower_l,lower_h],c='r',lw=1,linestyle='--')
+    plt.title(f"point estimate for {code}", fontsize=18)
+    plt.xlim(0,3)
+    plt.ylim(0,3)
+    plt.savefig(outfile, format='jpg')
+
+
+class EvaluatePointStats(object):
```
aimalz: This should be in `rail.evaluation` rather than the Golden Spike demo.

sschmidt23:
I looked through the goldenspike.py script yesterday and came up with a list of to-do's:
- [ ] refactor what goes into base.yaml (i.e. what is "hidden" from users by sticking it in a separate yaml file)
- [x] Update file locations for the outputs (right now some stuff just gets written directly to current directory, maybe move "results" specification to base.yaml to specify where things are saved. The "test" and "train" files that are designed to be inputs to the estimators also currently written to current directory, not ideal)
- [x] Implement a more generic spectroscopic degrader. Right now there is a specific degrader encoded, should replace with a way to specify a particular degrader and all options needed for that degrader
- [x] Remove GCRCatalogs?  It is of limited utility outside of cori, as you need the giant data files in addition to the gcd-catalogs package.  Now that I added a quick method to grab data from a local file, GCR seems unnecessary, as you can just assemble a file from a GCR query and use that.

sschmidt23:
added mention of alternate pip command for zsh, and added seaborn to evaluation's extras_require in `setup.py` (which also includes it in `[all]`

sschmidt23:
I finished all the updated I wanted, other than a refactor of base.yaml, which probably deserves its own issue after some discussion (a few options aren't really used practically when running batches of codes a la this goldenspike script).

Moving this from draft to ready to review, @eacharles comments would be welcome now that most of the temporary stuff has finally been removed.

examples/goldenspike/README.md
```diff
@@ -0,0 +1,49 @@
+# Golden Spike Example Script.
+
+"Golden spike" is meant to demonstrate a full "end-to-end" running of the RAIL infrastructure, something you might do to compare the performance of photo-z codes a la the PhotoZDC1 challenge described in [Schmidt et al. 2020](https://ui.adsabs.harvard.edu/abs/2020MNRAS.499.1587S/abstract).  There are three main modules that will be exercised within golden spike: mock data will be created with the Creation module, codes will be run with the Estimation module, and metrics computed with the Evaluation module.
+
+You must have RAIL and all dependencies installed locally in order to run goldenspike.  Follow the instructions on the main RAIL page (run `pip install .[all]`) in order to install `creation`, `estimation`, and `evaluation` RAIL packages.  In addition, if you will need either a pre-trained pzflow object or a local set of data with which to train the RAIL/creation creator that will create your mock data.  Note that there are both an example flow (`pretrained_flow.pkl`) and a small test set of data (`test_flow_data.pq`) available in the `RAIL/examples/goldenspike/data` subdirectory.
```
eacharles: The square brackets always have to be escaped in zsh (i.e., on new macs), so  you might just wanna put in a highlighted note to that effect somewhere here.

eacharles (approved):
Overall I think it makes sense to proceed.   

I got the script to run succesfully on sdf with no trouble whatsosever. 
On my mac I had a couple minor issues, the stuff about zsh, and the latex interpreter.  

I had a number of comments about next steps, but I think those a better addressed as new issues.   Let's get this merged in.

examples/goldenspike/README.md
sschmidt23: We added a note about escaping the '.[all]' on the main README.md, but I added another note here just in case.

PR #93, opened 2021-09-13
Update README.md

jfcrenshaw:
Quickly fixed a typo in the README installation instructions. 

Previously:
E.g. to install only the dependencies for the Creation Module, run `pip install .[creation]` or `pip install .[estimation]` respectively.

Now:
E.g. to install only the dependencies for the Creation Module **or the Estimation Module**, run `pip install .[creation]` or `pip install .[estimation]` respectively.

(the change is in bold)


aimalz (approved):

sjoudaki:
@jfcrenshaw @aimalz Please note that changes in the readme do not automatically propagate into docs/source (It’s the API documentation which is being auto-generated), so would need to update both.

jfcrenshaw:
Need to merge master back into this branch - that should fix the readthedocs failure (which is caused by outdated jax requirements in docs requirements.txt). I also need to copy the installation instructions into `docs/source/installation.rst`. I will do these today.

PR #96, opened 2021-09-24
Issue/35/error model

### LINKED ISSUE #35: Error model for the creation module
creator.py has the ability to include photometric errors using Eq 5 from https://arxiv.org/pdf/0805.2366.pdf
I need to finalize how this is handled, including flagging magnitudes beyond the 5-sigma limit.

Once that is done, I need to include error convolution in the posterior calculations. I have a way to do this analytically, and have it implemented, just need to finalize how to integrate it in the normalizing flow framework.

jfcrenshaw:
Implementing the LSSTErrorModel and incorporating changes made in pzflow 2.0.0

Still need to write unit tests. Starting the pull request so I can see coverage reports


rail/creation/degradation/lsst_error_model.py
```diff
@@ -0,0 +1,533 @@
...
+        In this case, you must include the additional information listed below...
+
+        IMPORTANT: For every band in bandNames, you must provide:
+            - nVisYr
+            - gamma
+            - the 5-sigma limiting magnitude. You can do this either by
+                (1) explicitly providing it in the m5 dictionary, or
+                (2) by adding the corresponding parameters to Cm, msky, theta,
+                and km, in which case the limiting magnitude will be calculated
+                for you, using Eq. 6 from the LSST Overview Paper.
```
sschmidt23, line 72: Maybe add an arxiv number for the overview paper that is the source of this model for anyone who might use it and be unfamiliar with the reference

rail/creation/degradation/lsst_error_model.py
```diff
@@ -0,0 +1,533 @@
...
+        self.m5 = self._calculate_m5()
+
+        # update the limiting magnitudes with any m5s passed
+        if m5 is not None:
+            self.m5.update(m5)
+
+    def default_settings(self):
+        """
+        Default settings for the error model. All of these values come from
+        the LSST Overview Paper, https://arxiv.org/abs/0805.2366
```
sschmidt23, line 183: maybe copy this link above to the first mention of the overview paper

rail/creation/degradation/lsst_error_model.py
```diff
@@ -0,0 +1,533 @@
...
+            Exposure time for a single visit
+        nYrObs : float, optional
+            Number of years of observations
+        nVisYr : dict, optional
+            Mean number of visits per year in each band
+        gamma : dict, optional
+            A band dependent parameter defined in the LSST Overview Paper
+        airmass : float, optional
+            The fiducial airmass
+        extendedSource : float, optional
```
sschmidt23, line 99: I think that this is a fudge factor that we added after the fact, and it might need a bit more explanation.  Maybe add a sentence like "The error model is meant to emulate the magnitude errors for point sources, if you wish to approximate the effect of extended source fluxes having larger uncertainties, a zeroth order correction is a simple magnitude offset for small, but slightly extended, sources.  `extendedSource` encodes that offset and which might be of order ~0.3 mags for typical LSST galaxies".

rail/creation/degradation/lsst_error_model.py
```diff
@@ -0,0 +1,533 @@
...
+        rng = np.random.default_rng(seed)
+        obsFluxes = rng.normal(loc=fluxes, scale=fluxErrs)
+
+        # only fluxes above minFlux will have observed magnitudes recorded
+        # everything else is marked as a non-detection
+        minFlux = 10 ** (self.settings["magLim"] / -2.5)
+        idx = np.where(obsFluxes > minFlux)
+
+        # convert fluxes back to magnitudes
+        obsMags = np.full(obsFluxes.shape, 99.0)
```
sschmidt23: We should check what DMStack does for magnitudes with negative/smaller than minFlux magnitudes.  Putting "99" is something that we did internally, as it's common in photo-z circles, but if we want to emulate the Stack outputs we should go with whatever they have and then make the photo-z codes deal with it internally.  The Stack might just leave these as `NaN`, I think.

rail/creation/degradation/lsst_error_model.py
jfcrenshaw: I include the link at the very start of the documentation for the class:

https://github.com/LSSTDESC/RAIL/blob/422a497547a0d0e5d15d434271c43034cddeb55b/rail/creation/degradation/lsst_error_model.py#L9-L13

But I agree it might be helpful to include it at the top of the `__init__` docstring too. And I'll make sure I refer to it as the LSST Overview Paper so that all references to that name are clear.

rail/creation/degradation/lsst_error_model.py
jfcrenshaw: Is there someone on Slack I can message about this?

rail/creation/degradation/lsst_error_model.py
jfcrenshaw: I will add this to the docstring. But we want to leave it defaulted to zero right?

rail/creation/degradation/lsst_error_model.py
jfcrenshaw: I clarified the name of the paper at the top of the class:
https://github.com/LSSTDESC/RAIL/blob/d06a767106affe2c35b5a5716dd4ff87e30632a9/rail/creation/degradation/lsst_error_model.py#L9-L14

And added a reference to the paper at the top of the `__init__` docstring:
https://github.com/LSSTDESC/RAIL/blob/d06a767106affe2c35b5a5716dd4ff87e30632a9/rail/creation/degradation/lsst_error_model.py#L24-L45

rail/creation/degradation/lsst_error_model.py
jfcrenshaw: [Resolved below](https://github.com/LSSTDESC/RAIL/pull/96#discussion_r734758670)

sschmidt23 (approved):
Everything looks good, checked that model agrees with the old implementation, minor changes to add a couple comments to doc, even if a couple bits of info get duplicated in multiple places. 

Only real comment is on the 99 vs NaN and checking what DMStack outputs.  I'll look into that.

rail/creation/degradation/lsst_error_model.py
sschmidt23: Yes, definitely want it default to zero.  A constant isn't a super good approximation of extended source effects; *but*, since steep number counts mean that most objects are faint and not at low redshift, many of them will be pretty small, and thus an approximate small bump to the m5/Cm value will be ok for most of those small & faint objects.  It's a terrible approximation for larger objects, though.  Best to leave the default behavior for point sources, as that's what the model is intended for originally.

rail/creation/degradation/lsst_error_model.py
sschmidt23: I think the most obvious thing to do is see what we have in GCRCatalogs for run2.2i, as that was run through DMStack.  Though we will need to be careful about what post-processing is done on the catalog before it's served to the user (i.e. look at the GCRCatalogs classes and see if they are doing any transforms on the data).  

Ultimately, we need to decide in DESC what our magnitude format will be.

rail/creation/degradation/lsst_error_model.py
jfcrenshaw: I updated the `__init__` docstring with a longer explanation:
https://github.com/LSSTDESC/RAIL/blob/a1f7fc86acb1882e1eb1bcd4b20c46ed764f2780/rail/creation/degradation/lsst_error_model.py#L98-L104

rail/creation/degradation/lsst_error_model.py
sschmidt23: I just checked GCRCatalogs, it takes fluxes and converts to nJy, and then converts that to AB mags with no checks for negative flux, which is why if you e.g. query `dc2_object_run2.2i_dr6_v2` for `mag_{band}_cModel` you almost always get a warning popping up when the query runs of 
```
/usr/local/py/envs/desc/lib/python3.8/site-packages/GCRCatalogs/dc2_dm_catalog.py:43: RuntimeWarning: divide by zero encountered in log10 return -2.5 * np.log10(flux) + AB_mag_zp_wrt_nanoJansky
```
I ran a query for a single tract on the `dc2_object_run2.2i_dr6_v2` catalog, and you do, indeed, get some `NaN` values for mag_cModel quantities, and those do correspond to objects with negative fluxes.

sschmidt23: I think PhotoZ is going to be one of the main "consumers" in DESC of the magnitudes in bands other than i-band, so we could have some influence on this.  If we wanted some post-processing to put values other than NaN for values below some threshold, I bet we could get the behavior changed for the DESC-specific catalogs, though.

jfcrenshaw: I decided to switch the default flag to NaN.

But I also added an option where the user can specify whatever flag they like:
https://github.com/LSSTDESC/RAIL/blob/c3d59d2106e97d16d2c5c9a0c84803cfc8538eb9/rail/creation/degradation/lsst_error_model.py#L110-L112

Issue #47, opened 2021-01-13
Input data naming in estimation codes


sschmidt23:
Currently the estimation codes assume that the ingested data has names "mag_{band}_lsst and mag_err_{band}_lsst for band in [u,g,r,i,z,y], and redshift as 'redshift' in the training data.  We should change this to a more generic naming for added flexibility.

johannct:
it would live in the config then? any idea what kind of generic naming could be used? Another point at stake is what to do if an algo is able to use ancillary information in the templating or learning, like the errors to start with?

sschmidt23:
The errors are already included, but yes, algorithms that want to use additional information is one concern, as well as photometry names that may have different names, e.g. including a "dered" or other extension in the name.

aimalz:
I agree that we'll have to standardize the naming/ordering of columns in all data we provide for each set of test conditions as well as give estimators the opportunity to establish the correspondence between those names and whatever variable names their code uses internally, while also minimizing the number of files that have to be updated if we have to make changes to the schema. Maybe we could keep hardcoded column names/orders out of the subclass modules by taking advantage of the dual global vs. per-estimator config structure? (That still wouldn't address if the meaning of the inputs changed, but I'm having trouble imagining an example that we couldn't preprocess our way around.)

There would be a bigger problem if the codes were expecting inherently different information, which we anticipate as the included codes grow in complexity. (I think @jmccull had put some thought into this early on.) Though we don't yet have a way to forward-model mock data consistent with SED templates (rather than the much simpler photometry necessary for machine learning codes), @sylvielsstfr is implementing the first (hybrid) template-fitting estimator subclass that will certainly encounter this in #49, and @sschmidt23 has mentioned wanting to wrap another soon. As those develop, I'm sure we'll gain a better understanding of the needs for the provision of shared ancillary data across codes that may be expecting diverse formats, let alone different content entirely.

johannct:
add LePhare++ to the list.....

jfcrenshaw:
This issue is also relevant to the Creation Module, where we need to decide on a default naming scheme for redshift and the LSST bandpasses in `Creator`, `FlowEngine`, and `LSSTErrorModel`.

My main request is that parameter errors are named `..._err`, as this is what pzflow looks for when looking for errors. For example, when looking for errors for a column named `u`, pzflow looks for a column named `u_err`.

yanzastro:
[...]

PR #105, opened 2021-10-26
Use `qp` in Creation Module, add notebook for posterior calculation, fix small errors in LSSTErrorModel

### LINKED ISSUE #99: Return pdfs from Creation Module as a qp object

jfcrenshaw:
The Creation Module now uses `qp` for posterior storage. I also removed the option to include posteriors with the sample, since this makes usage more clear and posteriors aren't stored in the DataFrame anymore. I also abstracted the `Creator` object a little more to keep all the pzflow-specific stuff in `FlowEngine`.

I added a new example notebook showing how to calculate posteriors, including the intricacies of using pzflow to calculate posteriors with error convolution and band marginalization.

Finally, I fixed a small error in `LSSTErrorModel` where if you set `ndFlag` to an integer, e.g. `ndFlag=99`, it would force the entire DataFrame to be of type int, which resulted in e.g. a bunch of 0 errors, which were just truncated floats < 1. I also one more check to the parameter validation to make sure the dictionaries of numbers all contain numbers, and are positive when appropriate


tests/creation/test_degraders.py
```diff
@@ -76,22 +83,30 @@ def test_random_seed(degrader, data):
...
     }
     degrader = BandCut(cuts)
+
     degraded_data = degrader(data)
+
     assert degraded_data.shape == data.query("u < 0 & y > 1 & y < 2").shape
 
 
 def test_BandCut_repr_is_string():
+    """Check that __repr__ returns a string
```
sschmidt23: For the future, for stuff like this that is pretty clearly trivial and exceedingly unlikely to really benefit from regular testing, I think we're all fine in adding a `#pragma: no cover` to the associated code block and skipping the simple test while still reporting 100% coverage.

sschmidt23 (changes requested):
Looks good, demo notebooks run.  A couple  minor comments that may or may not require changes, though we should discuss:

-I've been clearing jupyter outputs before committing to the repo just to save some space and keep things a little more clean looking.  Maybe clear the notebook, but generate a rendered html version and include that as well, as Julia did for the `evaluation` demo?

-Given that BandCut can be used to cut on any quantity, should it maybe be renamed to something like QuantityCut or something?  You state it in the demo and the docstring, but people are lazy, and having it called BandCut when it's actually more generic might curtail some potential usage by people that don't read said docstring

-Given that we do expect a population of non-detects in LSST mock data, do you think that we should maybe include some dedicated functionality for non-detect marginalization?  It might be done commonly enough that just having something prewritten that can be called would save some time for those generating the posteriors (I think we can assume that we'll only need to marginalize over one band, and most of the time that band will be u-band, which might simplify things a bit).

tests/creation/test_degraders.py
jfcrenshaw: Cool I didn't know about that command. I have removed the tests for `__repr__` methods and replaced them with with `#pragma: no cover`

jfcrenshaw:
> Maybe clear the notebook, but generate a rendered html version and include that as well

Done.

> Given that BandCut can be used to cut on any quantity, should it maybe be renamed to something like QuantityCut

I agree this is a good idea. I will make that change today

> do you think that we should maybe include some dedicated functionality for non-detect marginalization?

I agree in principle. I think we should discuss details during the standup today.

jfcrenshaw:
I added a default grid for u band marginalization, after checking that it works well for Y1 and Y10 errors. 

Note by default it assumes that the u band is stored as `u` because this is what the u band is named in my demos. However, by default, `LSSTErrorModel` assumes the u band is stored as `lsst_u`, because this is what it was called in the DC1 code. Perhaps we should agree on a default names for the LSST bands, and use these defaults everywhere?

sschmidt23:
I agree on the default names, and as a further example, the current estimation codes assume the names are 'mag_{band}_lsst', so some sensible internal defaults will minimize multiple name transforms (note: there is already an issue for this particular one: https://github.com/LSSTDESC/RAIL/issues/47).  Let's create a separate issue for that, or build off of issue #47 and not include it in this PR.

The only thing to do before merging is to clear the outputs on the posterior-demo.ipynb notebook, then I think this is ready to go.

jfcrenshaw:
I will add a comment about the default naming scheme to [issue 47](https://github.com/LSSTDESC/RAIL/issues/47), and ask that it be generalized to setting naming schemes RAIL-wide



> The only thing to do before merging is to clear the outputs on the posterior-demo.ipynb notebook

Whoops, nice catch! That notebook has now been cleared. Assuming all of the checks pass this last commit, this pull request is ready to be merged.

sschmidt23 (approved):

PR #109, opened 2021-11-01
add additional install instruction

sschmidt23:
After some problems with installing RAIL were reported by a user, Shahab and Markus asked me to expand the installation instructions a bit, this PR just adds some text to the README.md covering what to do if qp, DESC_BPZ, and FlexCode do not install correctly.


eacharles (changes requested):
Overall looks good.   Just two minor things:

pip install -e .[all] -> does not work on zsh (need quotes on '.[all]'

The examples depend on seaborn.

eacharles (approved):

PR #111, opened 2021-11-03
More accurate error model

### LINKED ISSUE #110: More accurate error model
The error model used in [Ivezic 2019](https://arxiv.org/pdf/0805.2366.pdf) is a high SNR approximation. This results in some weird behavior for low SNR galaxies. For RAIL, we want to use simulations as accurate as possible - so instead of using the high SNR approximation as DM does, we will use the more exact version of the error model. However, I will implement a flag `highSNRapprox` which will allow the user to use the high SNR approxiation instead.

Here I will explain the more accurate error model:

Eq. 5 from [Ivezic 2019](https://arxiv.org/pdf/0805.2366.pdf) states
*sig_rand^2 = (0.04 - gamma)\*x + gamma\*x (mag^2)*
however, instead of the square of the random error on the magnitude, this should be understood as *(N/S)^2*, where *N* and *S* are respectively the noise and signal in flux.

They equate this to a magnitude error because if you consider the observed magnitude *m_obs* to be composed of a true magnitude *m_true* and a magnitude error *m_err*, then you can write
*m_obs = m_true + m_err = -2.5 * log10(S + N) = -2.5 * log10[S(1 + N/S)] = -2.5 * log10(S) - 2.5 * log10(1 + N/S)*
and can therefore equate |*m_err*| *=  2.5 * log10(1 + N/S)*

They make the further approximation
*2.5 * log10(1 + N/S) = 2.5 / log(10) * N / S ~  1.0857 * N / S ~ N / S*,
where they drop the factor of 1.0857, as doing so actually improves the approximation in the low N/S regime.

Rather than making this approximation to calculate a magnitude error, I will use 
*N^2 = S^2 * [(0.04 - gamma)\*x + gamma\*x (mag^2)]*
to directly calculate more accurate flux errors, and then apply the errors as Gaussian in flux space.

jfcrenshaw:
This version of the error model is more accurate, not making the high SNR assumptions of the previous implementation. However, there is a flag `highSNR` which turns on all of those assumptions, thereby exactly replicating the Ivezic 2019 error model.

I also changed the default names of the bands in `LSSTErrorModel` to match the default names of the bands in the rest of the Creation Module. This is so that the Creation Module has a consistent default naming scheme. This makes it easier to use, and will make it easier to transition to whatever defaults we decide on for RAIL as a whole


jfcrenshaw:
Also fixed an edge case that broke `LSSTErrorModel` - specifically when you try to build an error model that adds errors for only a subset of the LSST bands.

sschmidt23 (approved):
Looks good, and behaves much more appropriately for low S/N (i.e. galaxies with S/N below the sky background end up with "observed" magnitudes that are ~1 sigma sky with appropriate errors, rather than potentially super bright).

PR #114, opened 2021-11-04
Added function to calculate limiting magnitudes.

### LINKED ISSUE #112: Add get_limiting_mags method to LSSTErrorModel
It would be useful for `LSSTErrorModel` to have a method to return N-sigma limiting magnitudes (both single exposure and coadded)

jfcrenshaw:


rail/creation/degradation/lsst_error_model.py
```diff
@@ -536,6 +534,63 @@ def __call__(self, data: pd.DataFrame, seed: int = None) -> pd.DataFrame:
...
+            nVisYr = np.array([self.settings["nVisYr"][band] for band in bands])
+            nStackedObs = nVisYr * self.settings["nYrObs"]
+        else:
+            nStackedObs = 1
+
+        # get the irreducible system error
+        if self.settings["highSNR"]:
+            nsrSys = self.settings["sigmaSys"]
+        else:
+            nsrSys = 100 ** (self.settings["sigmaSys"] / 2.5) - 1
```
sschmidt23: should this be 10 and not 100?

sschmidt23 (approved):
This looks good, but I think there is one typo (100 -> 10 in the one comment I left), but otherwise ready to merge.  I'll approve now assuming that you'll fix that, if it is indeed a typo

rail/creation/degradation/lsst_error_model.py
jfcrenshaw: Yes! I'll fix that now.

jfcrenshaw: I fixed it and added some more tests to make sure that the method is producing correct values over a range of `Nsigma` values:
https://github.com/LSSTDESC/RAIL/blob/70233c7f7cd65ea6a15a870d2799f05402662e6f/tests/creation/test_degraders.py#L197-L202

PR #116, opened 2021-11-24
Issue/94/pzflow estimator addresses #94

### LINKED ISSUE #94: Add a pzflow photo-z estimator to the Estimation Module
A Flow from pzflow can be used to estimate photo-z's. At some point, I want to add one of these to the Estimation Module, but it is currently low-priority.

sschmidt23:
This PR adds a basic `pzflow` based estimator to RAIL, and it seems to work well.  The only remaining to-do item is to add proper non-detection marginalization, this current implementation simply replaces training and test non-detects with a 1 sigma error limit for the magnitude and 0.75257 for the magnitude error.  Results look very good for my usual test data sets, with PZFlowPDF performing similar to, or better, than FlexZBoost.

I have not tested on a GPU machine, or at NERSC, I will do that now, but I don't foresee any differences from my usual setup.

One issue: the results on the very small dataset run in the tests differ between python3.7 and python3.8, so I have commented out the `assert np.isclose( [expected results], [calculated results])` test that I have for the other estimators.  There may be a seed not being set somewhere, or there may be a small difference in some jax library or something between python 3.7 and python 3.8.  Something to look in to for sure.


sschmidt23:
Oh, also: I had an exception from Jax from pzflow (I think it was the jax version of np.delete not being included) before I updated to a newer version of Jax.  So, pzflow may need a minimum jax version in the setup.py.

sschmidt23:
The code coverage reminded me that I forgot to add a test for the error marginalization option that I just added.  I just added the test, however, there is now a failure because I run a test that computes the mode values after an initial run, then re-run the estimation using the saved to file inform stage and compare the mode values to the initial mode values.  Again, there must be a seed or something different, as the mode values on the second run through do not agree with the first run.  Any ideas on how to make this repeatable, @jfcrenshaw ? 

(Edit: specifically, look at test_algos.py line 152:
`assert np.isclose(pz_dict['zmode'], rerun_pz_dict['zmode']).all()`
which is what is now failing, as the second pass modes are different from the first pass.

jfcrenshaw:
Re: the reproducibility problems, do you know if this only a problem when computing photo-z's with error marginalization? If so, it's because you need to add `seed=...` to the following call:
https://github.com/LSSTDESC/RAIL/blob/9c04c22b4055056efc4dcb70ecedcc76ea3bc42a/rail/estimation/algos/pzflow.py#L215-L217

sschmidt23:
I believe it occurs in the test with no error marginalization as well, though I think I'm also missing a seed argument there (i.e. I did not realize that posterior took a seed argument).  Let me try adding a seed in both cases and see if that resolves the problem.

jfcrenshaw:
> I believe it occurs in the test with no error marginalization as well, though I think I'm also missing a seed argument there (i.e. I did not realize that posterior took a seed argument). Let me try adding a seed in both cases and see if that resolves the problem.

Well the posterior method doesn't use the seed argument unless you are marginalizing errors

sschmidt23:
oh, hmm, I just pushed a change where I added seed to both. We'll see if it works.  Actually, hmm, we may need to add zb_expected to the pytest decorator with different values if they change from the mode values without the error marginalization.

sschmidt23:
It looks like the mode estimates are bouncing around a little bit when I re-run pytest on my local machine both with and without the error marginalization.  Can you run the tests and see if you see the same stochasticity?

jfcrenshaw:
I have to log off for the day, so I probably won't test this for a few days. But I did just run this test on my computer:
```
import numpy as np
from pzflow import Flow
from pzflow.bijectors import RollingSplineCoupling

grid = np.linspace(-2, 2, 10)

flow1 = Flow(["x", "y"], RollingSplineCoupling(2))
s1 = flow1.sample(10, seed=0)
p1 = flow1.posterior(s1, "x", grid)
flow1.save("test.pkl")

flow2 = Flow(file="test.pkl")
s2 = flow2.sample(10, seed=0)
p2 = flow2.posterior(s2, "x", grid)

assert s1.equals(s2)
assert np.allclose(p1, p2)
```
and it passed, so at the very least I think it's an error in the estimator and/or test and not in pzflow itself

sschmidt23:
I do run inform in the test, so there is a flow.train in the test.  My guess is that the differences are arising as the loss is minimized using the training data.

jfcrenshaw:
Actually got a chance to look at this again this morning. It looks like you fixed the issue?

Also re: speeding up the test, I thought it was probably slow because pzflow does a lot of just-in-time (jit) compilation. That speeds up training, but adds overhead time upfront. Since this is such a small training set with so few epochs, I thought disabling jit would speed it up. I tried adding
```
with jax.disable_jit():
    pz_dict, rerun_pz_dict = one_algo(pz_algo, config_dict)
```
but I think this actually made it slower.

Unfortunately I think speeding up these tests would require some code optimization for pzflow. I have already been thinking about doing some refactoring and optimization for pzflow, but it's not high on my priority list at the moment. 

But otherwise I think this looks good. I'll go ahead and approve it. Let me know if there's anything else you want me to take a look at before we merge it.

jfcrenshaw (approved):

sschmidt23:
I didn't "fix" it per se, but just increased atop on the np.isclose test.  But, I think I'm ok with a small jitter in the mode value due to slightly different loss minimization setups for python 3.7 vs 3.8.  So, I'm going to add a little explanation of the config params in the rail/estimation/algos/README.md, then I'll merge this in.

PR #119, opened 2021-12-01
add params to goldspike yaml that won't have it run super slow

sschmidt23:
The golden spike script did not have include_mag_errors or n_error_samples keywords (I forgot to add to estimation_goldspike.yaml when I updated the pzflow algos file).  Having explicit keywords in the example config might be useful for someone doing their first explorations, so I'm adding them real quick in this trivial PR, which I'll merge myself.


sschmidt23:
damn, I don't have permission to force-merge without a review, @eacharles or @jfcrenshaw can you approve this PR real quick?

jfcrenshaw (approved):
Looks good. I checked that these new params get used in `rail/estimation/algos/pzflow.py` too.

PR #127, opened 2022-03-16
Renaming: "train" --> "inform"

aimalz:
The tests pass on my local machine, but I'm making this mini-PR to make sure I'm not breaking anything for GA either.


eacharles (changes requested):
Can you get the notebooks as well please?

Thanks!
-e

PR #133, opened 2022-04-07
fix issues in goldenspike nb

sschmidt23:
I made some updates to the notebook, there were multiple bits commented out, e.g. all of the actual evaluation stuff, and for some reason incompleteness was added to the test data rather than the train data.  I also swapped out the less interesting randomPZ, trainZ, and simpleNN for KNearNeighPDF and FlexZBoost.

This now runs fine on my local machine, and runs on cori *if* you set uselatex to false in up's qp/plotting.py (or if you set up LaTex on cori before you run).

### DIFF
--- a/examples/goldenspike/goldenspike.ipynb
+++ b/examples/goldenspike/goldenspike.ipynb
@@ -2,7 +2,6 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "449c685e",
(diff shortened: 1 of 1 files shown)


eacharles (approved):
Looks good.   Sorry for breaking things.   I had misunderstood the reasons for having the test and training samples differ.  

Thanks!

PR #142, opened 2022-04-13
Issue/141/simplereadme addresses #141

### LINKED ISSUE #141: Simplify README.md
We have a  lot of duplicate info in the README.md that is also on the readthedocs page.  For ease of maintaining this and to avoid these getting out of sync, we should eliminate most of the duplicated text from README.md and simply point to the readthedocs page for installation and usage.

sschmidt23:
Addresses issue #141 
Remove a lot of the duplicated text from README.md and instead just point to the Read The Docs page. Also, add the references to specific codes for RAIL that should be cited if used to the citing.rst section of the read the docs.


eacharles (approved):
Thanks Sam.

sschmidt23:
I just noticed the links are in the wrong format on the installation.rst page, I'm just going to fix those in this PR before merging rather than create a separate issue, just need to switch from the copied .md to .rst syntax for links.

PR #144, opened 2022-04-13
Issue/143/readthedocs

eacharles:
At least got it to populate the API, e.g.,

https://lsstdescrail.readthedocs.io/en/issue-143-readthedocs/api.html


sschmidt23:
Does the fact that it's not finding the docstrings mean that we are incomplete in our docs/requirements.txt?  It sounds like readthedocs actually has to have everything install and then uses python to actually grab the docstrings, so if we're missing dependencies then that might be why it's failing.  I think we are a bit incomplete there, I was just looking through the setup.py and checking which packages we don't have listed, I'll update in a minute, and we can see if that fixes things.

sschmidt23 (approved):
Ok, I think this looks good now, other than Delight and its annoying cython dependency, which we can track down later

PR #147, opened 2022-04-14
Added stuff for dynamic versioning

eacharles:


eacharles:
I removed the .readthedocs.yaml, it was way out of date and causing problems (basically it didn't tell readthedocs to call `python setup.py install`, so the _version.py file wasn't getting set. 
For now we are just using the default that is configured from this page:
 https://readthedocs.org/dashboard/lsstdescrail/edit/

sschmidt23 (approved):
Works, approved

PR #149, opened 2022-04-14
move data files and modify defaults for bpz and delight SED/FILTER

sschmidt23:
Just moves the AB, SED, and FILTER directories from rail/estimation/data to examples/estimation/data


aimalz (approved):
Looks great, thanks!

PR #152, opened 2022-04-15
moving summarization and propagating to tests

aimalz:
The tests work but the golden spike notebook is broken because of absolute paths in the ceci configuration.


eacharles:
Which paths?  In which configuration file?

eacharles (approved):
Ok, I've picked up the recent changes to main, and done a bit of delinting.  I think this is good to go.

PR #155, opened 2022-04-15
added render_nb.sh

eacharles:


sschmidt23 (approved):
I'm marking as approved, test is failing, but that's the unrelated new qp warning for mixmod.

But, we should add the new notebook from PR #159 (that I'm just about to merge in) to be rendered by the script as well before merging.

PR #156, opened 2022-04-15
Issue/153/nondetect

### LINKED ISSUE #153: handling of non-detects
Several of the algorithms have hardcoded value assumptions for non-detections, we should change so that this is flexible, and includes possibility of np.nan.

sschmidt23:
fix the non detect issue, setting as a draft until I see that all the tests pass and I didn't break anything while fixing.


eacharles (approved):
Looks good.  Thanks.

PR #160, opened 2022-04-19
more conda doc stuff

sschmidt23:
There was already a little bit of mention of a fresh conda environment, but I made it a little more explicit, and also explicitly mentioned some caveats like python3.7 not supporting the newest scipy, and added the commands to add an ipython kernel from your new env, which is useful for people running at NERSC.


eacharles (approved):
Thanks Sam!

PR #162, opened 2022-04-20
Issue/161/testskip

sschmidt23:
This PR just adds a test skip for a scipy<1.8.0 that we know should fail for the KNearNeighPDF test that uses mixmod from qp, and also puts a (temporary) limit on the pzflow version in setup.py that will keep tests passing until John Franklin updates RAIL to reflect recent changes in pzflow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ name: rail
 
 on:
   push:
-    branches: [ issue/161/testskip ]
+    branches: [ main ]
   pull_request:
     branches: [ main ]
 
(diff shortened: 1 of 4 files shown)


sschmidt23:
Ah, dammit, I have two other commits that update the docs stuff in here as well from the other PR that I was working on, sorry.  They just add a few more sentences on the conda stuff and a mention of how to add your conda env as an ipykernal for use in a notebook (which can be useful for the NERSC Jupyter hub).

eacharles (approved):
Looks good

PR #166, opened 2022-04-21
remove inaccurate language

### LINKED ISSUE #164: degradation notebook needs to be updated
As Irene Moskowitz pointed out in the desc-pz-rail channel, the degradation-demo notebook says that the QuantityCut degrader will draw samples to replace those cut and return a dataframe of the same original size.  This was removed in the switch to ceci, we need to update the notebook and any documentation that makes this claim.

sschmidt23:
Trivial change that removes inaccurate language that confused someone, as degraders that remove objects no longer sample new objects to get to the same number of objects as the input.

### DIFF
--- a/examples/creation/degradation-demo.ipynb
+++ b/examples/creation/degradation-demo.ipynb
@@ -276,7 +276,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


eacharles (approved):
Thanks Sam.

PR #167, opened 2022-04-21
move summarization text to estimation, reword a bit

sschmidt23:
This now reflects the updated organization of RAIL, combines estimation and summarization in overview.rst


aimalz (approved):
I approved the PR but did some rearrangement and clarification while reviewing the changes so will leave the branch unmerged.

PR #170, opened 2022-04-28
U/sschmidt23/nzdir

### LINKED ISSUE #130: re-implement the TXPipe implementation of NZDir in RAIL
This is already roughly done, I just forgot to create an issue, the code is in the u/sschmidt23/nzdir branch, internals are a little rough but it's functional (see demo notebook in that branch in examples/summarization/NZDir.ipynb), I'll finish fixing this up once I am back from vacation/LINCC conference.

sschmidt23:
Porting the TXPipe implementation of the NZDir algorithm implemented here: https://github.com/LSSTDESC/TXPipe/blob/nz-dir/txpipe/nz_calibration.py
as it seems more appropriate for RAIL rather than TXPipe
@joezuntz does this meet all of the needs currently covered by the TXPipe version?  This works, and works well for representative data, however, the implementation shortcut of checking radii around spec-z objects rather than photometric objects means that any photometric objects that are not covered by the annuli around the N spec-z objects will not be counted/weighted in the N(z) sum, but will still be present in the bin, which can lead to bias in the estimate. 

examples/estimation/NZDir.ipynb notebook demonstrates use of the algorithm with representative and incomplete data, including bias in the high-z bin for incomplete data.
  
Addresses issue #130


rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+                          kmetric=Param(str, "euclidean", msg="Knn metric to use"),
+                          szname=Param(str, "redshift", msg="name of specz column in sz_data"),
+                          szweightcol=Param(str, "weight", msg="name of sz weight column"),
+                          distance_delta=Param(float, 1.e-6, msg="padding for distance calculation"),
+                          hdf5_groupname=Param(str, "photometry", msg="name of hdf5 group for data, if None, then set to ''"))
+
+    def __init__(self, args, comm=None):
+        """ Constructor:
+        Do Informer specific initialization """
+        CatInformer.__init__(self, args, comm=comm)
```
joezuntz, line 46: As of py3 you can just do `super().__init__` for this, which helps if you change the class name later.

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+
+    bands = ['u', 'g', 'r', 'i', 'z', 'y']
+    default_usecols = [f"mag_{band}_lsst" for band in bands]
+
+    name = 'NZDir'
+    config_options = CatEstimator.config_options.copy()
+    config_options.update(zmin=Param(float, 0.0, msg="The minimum redshift of the z grid"),
+                          zmax=Param(float, 3.0, msg="The maximum redshift of the z grid"),
+                          nzbins=Param(int, 301, msg="The number of gridpoints in the z grid"),
+                          usecols=Param(list, default_usecols, msg="columns from sz_date for Neighor calculation"),
```
joezuntz, line 107: typo `sz_date`

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def run(self):
+        if self.config.hdf5_groupname:
+            sz_data = self.get_data('input')[self.config.hdf5_groupname]
+        else:  # pragma:  no cover
+            sz_data = self.get_data('input')
+        if self.config.szweightcol in sz_data.keys():  # pragma: no cover
+            szweights = np.array(sz_data[self.config.szweightcol])
+        else:
+            szweights = np.ones(len(sz_data[self.config.usecols[0]]))
```
joezuntz, line 57: I got a bit confused here because I thought the u-band was being used for the weights. My fault for being unobservant, but maybe add a comment explaining that it's unit weights?

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+                          hdf5_groupname=Param(str, "photometry", msg="name of hdf5 group for data, if None, then set to ''"))
+
+    def __init__(self, args, comm=None):
+        """ Constructor:
+        Do Informer specific initialization """
+        CatInformer.__init__(self, args, comm=comm)
+
+    def run(self):
+        if self.config.hdf5_groupname:
+            sz_data = self.get_data('input')[self.config.hdf5_groupname]
```
joezuntz, line 50: This is my ignorance of RAIL I think, but does the name "input" get replaced with something else somewhere? Possibly this is the new ceci feature that Eric added.

EDIT: Figured this out - this is the aliases feature, and it works with TXPipe

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+                          phot_weightcol=Param(str, "weight", msg="name of photometry weight, if present"))
+
+    def __init__(self, args, comm=None):
+        self.zgrid = None
+        self.model = None
+        self.distances = None
+        self.szusecols = None
+        self.szweights = None
+        self.sz_mag_data = None
+        CatEstimator.__init__(self, args, comm=comm)
```
joezuntz, line 119: Can also use `super()` here

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def open_model(self, **kwargs):
+        CatEstimator.open_model(self, **kwargs)
```
joezuntz, line 122: Can also use `super()` here

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+    def run(self):
+        if self.config.hdf5_groupname:
+            test_data = self.get_data('input')[self.config.hdf5_groupname]
+        else:  # pragma:  no cover
+            test_data = self.get_data('input')
+        self.zgrid = np.linspace(self.config.zmin, self.config.zmax, self.config.nzbins)
+
+        # assign weight vecs, if present, else set all to 1.0
+        # tested in example notebook, so just put a pragma no cover for if present
+        if self.config.phot_weightcol in test_data.keys():  # pragma: no cover
```
joezuntz: It looks like if you set `phot_weightcol` but it's not present in the data this will silently be ignored. Is that intended?

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+    def open_model(self, **kwargs):
+        CatEstimator.open_model(self, **kwargs)
+        self.distances = self.model['distances']
+        self.szusecols = self.model['szusecols']
+        self.szweights = self.model['szweights']
+        self.szvec = self.model['szvec']
+        self.sz_mag_data = self.model['sz_mag_data']
+
+    def run(self):
+        if self.config.hdf5_groupname:
```
joezuntz, line 130: This is the bit that I think changes later when Josue's table_io changes come through. (no change needed now, just thinking through this)

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+        self.zgrid = np.linspace(self.config.zmin, self.config.zmax, self.config.nzbins)
+
+        # assign weight vecs, if present, else set all to 1.0
+        # tested in example notebook, so just put a pragma no cover for if present
+        if self.config.phot_weightcol in test_data.keys():  # pragma: no cover
+            pweight = np.array(test_data[self.config.phot_weightcol])
+        else:
+            pweight = np.ones(len(test_data[self.config.usecols[0]]))
+        # calculate weights for the test data
+        if self.config.bincol not in test_data.keys():
```
joezuntz: It would be useful for TXPipe to be able to get the bin col from one file and the photometry data from another (because we will want to try various binnings for the same photometry). I'm not sure if that's possible here, maybe by doing something with the data store?  Maybe this is a TODO.

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+            # stack all galaxies into single ensemble
+            print("bincol not found! Perform single tomo bin measurement!")
+            tomobinids = np.ones(len(test_data[self.config.usecols[0]]), dtype='int')
+            test_data[self.config.bincol] = tomobinids
+        else:  # pragma: no cover
+            print("found bin IDs")
+            tomobinids = test_data[self.config.bincol]
+            print(tomobinids)
```
joezuntz: Probably want to remove this printout for laterger runs!

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+        # away to search a second tree of photo data
+        distances, _ = neighbors.kneighbors(sz_mag_data)
+        distances = np.array(np.amax(distances, axis=1) + self.config["distance_delta"])
+        self.model = dict(distances=distances,
+                          szusecols=self.config.usecols, szweights=szweights,
+                          szvec=szvec, sz_mag_data=sz_mag_data)
+        self.add_data('model', self.model)
+
+
+class NZDir(CatEstimator):
```
joezuntz, line 80: Is this a summarizer rather than an informer?  Not sure if it makes a difference; maybe they end up the same.

rail/estimation/algos/NZDir.py
```diff
@@ -0,0 +1,182 @@
...
+
+            # for each of the indexed galaxies within the distance, add the weights to the
+            # appropriate tomographic bin
+            for j, index in enumerate(indices):
+                weights[i, j] += pweight[index].sum()
+            # make weighted histograms
+            single_hist = np.histogram(
+                self.szvec,
+                bins=self.zgrid,
+                range=(0, self.config["zmax"]),
```
joezuntz: I don't think `range` does anything if you set `bins` as well?

rail/estimation/algos/NZDir.py
sschmidt23: There's been a bit of back and forth about the terminology that we use, and what exactly a "summarizer" is.  I think the confusion comes up because we have some tomographic bin estimates which operate using the ensemble of individual galaxy PDFs (e.g the "naive stack" and "point estimate histogram" summarizers), but we also have others, like this NZDir tomographic bin estimate, that do not estimate individual PDFs at all.  I *think* we decided that we can, instead, refer to these latter types of algorithms as "N(z) estimators".  Also, when I decided to split up the "training" of running the initial spec-z tree to determine distance neighborhoods as being separate from the ball tree finding of photometric neighbors, the class structure of catInformer and catEstimator just fit better than the Summarizer classes.  Note that I originally wrote this as a Summarizer class, you can see (look at the commit just before the "switch to inform/estimate rather than summarize" one if you are curious), but felt that this would work better for eventual parallelization, as the checking of photometry neighbors around the spec-z objects should be trivial to run in parallel and then stack up at the end.

rail/estimation/algos/NZDir.py
sschmidt23: Yes, once Josue's changes are in place we can modify this to check chunks of photometry matches within the distance annuli around the spec-z objects and then add the histogram values of the chunks when we finalize.  We'll deal with that once Josue's stuff is finalized, and just leave as-is for now.

rail/estimation/algos/NZDir.py
sschmidt23: super() causes an error for "multiple values of comm" due to how comm is used in the superclass/subclass, so I'll leave them as the specific class names for now, we can revisit that as a separate issue.

rail/estimation/algos/NZDir.py
joezuntz: Makes sense, thanks for explaining!

rail/estimation/algos/NZDir.py
sschmidt23: comment added, and I actually changed this to actually raise an error if the key is asked for but the column is not present in the data file

rail/estimation/algos/NZDir.py
sschmidt23: Following Alex's suggestion, I actually removed the multi-bin functionality, so the method now expects a single tomographic bin as input.

sschmidt23:
I removed the simultaneous estimation of multiple tomographic bins via the "bincol" column, so now the estimator expects you to make the cuts first and pass in one bin at a time.  I made appropriate changes to the demo notebook to reflect this, and all runs fine.  I also addressed all of Joe's other minor comments, other than the super().__init __, which is still complaining about comm for some reason.  If we feel like tracking that down then we can do it in a separate issue, as most/all of the estimators have the superclass name hardcoded rather than using super(), and so it is larger than just NZDir.

joezuntz (approved):

joezuntz:
Looks good!

PR #172, opened 2022-04-29
Issue/171/hsc degrader

imoskowitz:
Added a new class called HSCSelection to spectroscopic_degraders.py implementing HSC-like selection of training samples (description below). Also added a file, hsc_ratios.csv to the RAIL directory. Not sure where the best place to put that is.

- HSC galaxies are divided into pixels based on i-band magnitudes and g-z colors. In each pixel, I compute the ratio between HSC galaxies that have spectroscopic redshifts (training) to galaxies with only photometry (application). The ratios at each pixel are saved to a csv file.

- The degrader reads in the file with the HSC ratios, and divides galaxies into the same pixels. Within each pixel, a random subsample is selected based on the HSC ratio for that pixel (i.e if there are 10 galaxies in a pixel, and the ratio for that pixel is 0.4, 4 random galaxies will be selected from that pixel to pass into the training sample). This is the realistic case.

- The degrader has an optional parameter to specify a redshift cutoff, in which case galaxies selected for the training sample are further rejected if the redshift is above the cutoff.


hsc_ratios.csv
```diff
@@ -0,0 +1,200 @@
...,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00,0.000000000000000000e+00
```
sschmidt23: we might want to use hdf5 or parquet for this file rather than csv

rail/creation/degradation/spectroscopic_degraders.py
```diff
@@ -134,3 +134,91 @@ def run(self):
...
+        g_mag = data['g'].to_numpy()
+        z_mag = data['z'].to_numpy()
+
+        gz_color = g_mag-z_mag
+        
+        data['color'] = gz_color
+        
+        #now remove galaxies that don't fall in i-mag and g-z color range of HSC data. These will always be application sample galaxies
+        
+        data_hsc_like = data[(data['i'] >= 13) & (data['i'] <=26) & (data['color'] >= -2) & (data['color'] <=6)]
```
sschmidt23: Rather than hardcoding numbers, we might want to change these to config parameters that are read in with the bit HSC grid as part of a model (maybe in a pickled dictionary), that would allow for more flexibility if we update the parameters, and might make this more generally useful, e.g. we could write similar pickle file and use this same code for DEEP2.

rail/creation/degradation/spectroscopic_degraders.py
```diff
@@ -134,3 +134,91 @@ def run(self):
...
+            ratio_list.append(ratios[pixels_y[i]][pixels_x[i]])
+        
+
+        data_hsc_like['ratios'] = ratio_list
+
+        #This picks galaxies for the training set
+        unique_ratios = data_hsc_like['ratios'].unique()
+
+        keep_inds = []
+        for i in range(len(unique_ratios)):
```
sschmidt23: I think there's probably a more efficient way to do this double loop in numpy, though I'm bad at numpy array stuff, so I'll have to dig around a bit to remind myself what that better way is.

rail/creation/degradation/spectroscopic_degraders.py
```diff
@@ -134,3 +134,91 @@ def run(self):
...
+            temp_data = data_hsc_like[data_hsc_like['ratios'] == unique_ratios[i]]
+            number_to_keep = int(len(temp_data)*unique_ratios[i])
+            indices_to_list = temp_data.index.values
+            for j in range(0, number_to_keep): # pragma: no cover
+                keep_inds.append(indices_to_list[j])
+ 
+
+        training_data = data_hsc_like.loc[keep_inds,:]
+
+        #For the pessimistic choice, also remove galaxies with z > redshift_cut from the sample
```
sschmidt23: Rather than have this built-in, could we just apply the QuantityCut degrader after the fact?  This duplicates that functionality.  I guess I can see utility in having it built in for simplicity, so I can see either keeping or removing this.

rail/creation/degradation/spectroscopic_degraders.py
```diff
@@ -134,3 +134,91 @@ def run(self):
...
+
+        data_hsc_like['ratios'] = ratio_list
+
+        #This picks galaxies for the training set
+        unique_ratios = data_hsc_like['ratios'].unique()
+
+        keep_inds = []
+        for i in range(len(unique_ratios)):
+            temp_data = data_hsc_like[data_hsc_like['ratios'] == unique_ratios[i]]
+            number_to_keep = int(len(temp_data)*unique_ratios[i])
```
sschmidt23: This formulation is not ideal for low-occupation cells, as if you have a cell with only a few galaxies and a relatively small fraction, no galaxies from that cell will ever be chose.  I think a smarter thing to do is for each galaxy draw a np.random.uniform number and if that number is < ratio keep the galaxy, if not reject it.  That will allow at least a few of the "rare" galaxies to get chosen from cells with low ratio, particularly when viewing the collective bulk of low ratio cells.  That is, if you consider the collection of say 100 cells an area of color space where all of the ratios are ~0.05-0.1, all with 5 galaxies, the current formulation would choose zero objects in those 100 cells, while drawing 500 random numbers and comparing to the ratios would keep something like 25-50 galaxies in that 100 cell area.

sschmidt23 (changes requested):
This looks very nice, though I do think we should change the way we choose which galaxies to keep in L210 of spectroscopic_degraders.py to a comparison to a random draw.  Also, we can take some of the hard coded values and read those as config params.  With a few changes we can make this a little more general so that we can re-use the 2D ratio stuff for DEEP2 and other degraders as well.

sschmidt23 (approved):
I chatted with Irene, she and I made the appropriate changes, things now all look good.

PR #175, opened 2022-05-05
fix prior_band in goldenspike_config.yml, also, make error message mo…

eacharles:
…re informative


sschmidt23 (approved):
Looks good, thanks for fixing.

PR #188, opened 2022-05-18
Issue/187/tables io fixes

eacharles:
Fixes error introduce by changes to tables_io


joselotl (approved):
The changes look good. Sorry, I saw the write_chunk, but didn't realized that the upgrade in tables_io would break it.

Issue #72, opened 2021-06-22
`bpz_lite` "inform" option


sschmidt23:
The original BPZ code has no method to train a new prior, it is really only designed to be run with pre-computed and named priors, e.g hdfn_gen (based on very sparse HDF-N data).  Currently `inform` in `bpz_lite` is just `pass`.

We could implement a training procedure with the following steps:
1. fit the "best fit type" at the fixed spec-z for each training galaxy
2. break the individual SED types into "broad types", e.g. Elliptical, Spiral, SF
3. Fit parameterized prior of the form in Benitez (2000) for these broad types as a function of magnitude, type, and redshift
4. write parameters in such a way that they can be used in a function called from within BPZ

sschmidt23:
closed by #191

PR #191, opened 2022-06-02
Issue/72/bpz inform

sschmidt23:
Addressing issue #72, this PR adds an inform stage to train a prior for BPZ when given a set of "broad type" classifications for the training galaxies, and uses a newly added feature of [DESC_BPZ](https://github.com/LSSTDESC/DESC_BPZ) where rather than hardcoded values in a named prior, you can now supply a pickled dictionary of prior params to specify a prior.  
For now, the user must separately determine the "broad type" for each training galaxy, which is read in as a set of integers from file, there is not a great way to automate that process that I'm aware of, so any attempt to do so will be left to a future addition.

Inform is really a pretty straightforward likelihood minimizer that fits for the same parameterization described in the original Benitez (2000) BPZ paper and used in the default HDFN_gen prior.  It first fits for the p(broad type | apparent mag) for each broad type, then fits p(redshift | apparent mag, broad type) assuming a form z**a * exp(-z/zm)**a where zm = zo + km(m-mo), i.e. the peak of the redshift shifts linearly with apparent mag in a way that best matches the training data for each broad type.

Everything seems to work, prior looks like the training data from cosmoDC2 on a small test set, and the custom prior gives better metrics than the "default" that is not tuned to the data.


eacharles (approved):
Looks good.  Thanks!

PR #192, opened 2022-06-03
Move imports inside functions

joezuntz:
As an umbrella package that connects lots of different codes together, RAIL imports loads of differert python packages. This leads to slow import times, and requirements that lots of depdenencies are installed even if you don't plan to use them.

This PR moves the imports of many of the packages inside the functions and methods where they are used.  This means far less stuff is loaded when `import rail` happens and is instead deferred until needed.


johannct:
It does not seem at all like the proper "pythonic" way to do it, right? It would be good to understand the problem at a more fundamental level, I believe..... which boils down to the promotion of every modules occurring within the various __init__.py

joezuntz:
It's not incorrect in any way, but you're right that most style guides don't approve (though many company ones assume long-running programs where the start-up time isn't relevant).

We do want the different classes to be automatically searchable, and this is the only way I know to make that work while not importing everything automatically.

johannct:
why not add some logic in the __init__ file, doing promotion only on the basis of existing underlying libs? Seems cleaner to me, as you still do not want to have "automatic searchability" of a class you cannot use, right? So forbidding these when e.g. estimator builds its registry of possible estimator modules is still ok. Long time I did not look at the code, so maybe I am missing something, so it may be better that a second reviewer chimes in. But I stay with my hunch : likely not the right way to deal with the issue.

sschmidt23:
I had a similar reaction to Johann initially, in that this seems to not fit the usual python idiom; however, RAIL does have several dependencies that are slow to load, and this would speed them up, so it seems like a justifiable exception.  I was trying to think of some way to document this a bit, both package-wide and file by file (for those who might be digging through the code and glancing at the imports at the top of the file to get a rough idea of dependencies).

johannct:
@sschmidt23 I am not at all arguing about the usefulness of and need for a solution. I am arguing that the problem is the global promotion of all modules in the __init__ files. I understand and remember the issue with introspection, but I am almost certain that it is at the __init__ stage that some vetoing of promotion should occur, based on the availability of the corresponding dependencies.

joezuntz:
I disagree on that last point - I think we want the classes to be visible, even if the dependencies are not present.  For example, we can currently do `python -m rail` to list all available stages, and it's very handy to be able to see all the possible choices.

joezuntz:
It currently generates:
```
Usage: python -m <module_name> <stage_name> <stage_arguments>

If no stage_arguments are given then usage information
for the chosen stage will be given.

I currently know about these stages:

- ColumnMapper               - Utility stage that remaps the names of columns.
- RowSelector                - Utility Stage that sub-selects rows from a table by index
- TableConverter             - Utility stage that converts tables from one format to anothe
- QuantityCut                - Degrader that applies a cut to the given columns.
- LSSTErrorModel             - LSST Model for photometric errors.
- LineConfusion              - Degrader that simulates emission line confusion.
- InvRedshiftIncompleteness  - Degrader that simulates incompleteness with a selection function
- GridSelection              - Uses the ratio of HSC spectroscpic galaxies to photometric galaxies to portion a sample
- FlowEngine                 - Engine wrapper for a pzflow Flow object.
- FlowPosterior              - Engine wrapper for a pzflow Flow object
- Evaluator                  - Evalute the preformance of a photo-Z estimator
- Inform_PZFlowPdf           - Subclass to train a pzflow-based estimator
- PZFlowPDF                  - CatEstimator which uses PZFlow
- PointEstimateHist          - Summarizer which simply histograms a point estimate
- Inform_BPZ_lite            - Placeholder class for the eventual BPZ_lite inform stage
- BPZ_lite                   - CatEstimator subclass to implement basic marginalized PDF for BPZ
- Inform_FZBoost             - Train a FZBoost CatEstimator
- FZBoost                    - FZBoost-based CatEstimator
- RandomPZ                   - Random CatEstimator
- Inform_KNearNeighPDF       - Train a KNN-based estimator
- KNearNeighPDF              - KNN-based estimator
- VarInferenceStack          - Variational inference summarizer based on notebook created by Markus Rau
- Inform_SimpleNN            - Subclass to train a simple point estimate Neural Net photoz
- SimpleNN                   - Subclass to implement a simple point estimate Neural Net photoz
- Inform_trainZ              - Train an Estimator which returns a global PDF for all galaxies
- TrainZ                     - CatEstimator which returns a global PDF for all galaxies
- NaiveStack                 - Summarizer which simply histograms a point estimate
- Inform_NZDir               - Quick implementation of an NZ Estimator that
- NZDir                      - Quick implementation of a summarizer that creates
```

johannct:
@joezuntz I am not sure what the use is of showing "- FZBoost                    - FZBoost-based CatEstimator" to a user who can't use it. But nevermind, this is a documentation take at it, and documenting specific modules and their dependencies is another topic. What I would find natural in a session  is to do " from rail import * " and then have a registry of the available estimators given the found deps. The matter of letting the user know that some estimators might be absent because of dependencies unmet is certainly useful as well, but is a different problem. For what is worth I do not understand why you use the term "stage" for the list above :)  and I am not sure I find it super useful, especially showing utilities like ColumnMapper together with other things like the CatEstimators available. I presume it is a matter of taste :) . Anyway, I lost ground on the rail dev, and I am not ready to dive in again just yet (knee deep into lephare++) so the core developers should just go ahead with whatever solution they fancy best.

johannct:
I realize that there are actually 2 problems : dependencies to meet, and slow imports when they are met. These are two different problems no? I had essentially the first one in mind in the comments above.

joezuntz:
> I am not sure what the use is of showing "- FZBoost - FZBoost-based CatEstimator"

For me the utilties are:
- showing the user what other options they could be running if they install the deps
- letting the user design a pipeline on one machine and run it on another. I regularly do this between my laptop / NERSC.

> For what is worth I do not understand why you use the term "stage" for the list above

That's a ceci term, the idea is that all these classes are stages (or steps) in a pipeline, connected together. The base classes are `ceci.PipelineStage -> rail.RailStage`

I do agree these are two problems, yes. That's a good point.

johannct:
ok thanks @joezuntz I forgot about the ceci vocabulary. So we have 3 problems actually : the third one is proper and useful "runtime" documentation. I understand the way you use the list above now, and ok it can be useful. But it still is yet another separate issue. We are talking about foundational aspects of rail here, it may be a bit late, but we should make sure we have an expert take into how to design all this as well as possible. And I am definitely no expert!

eacharles:
Ok, my take on this that we should merge this PR because it will make thing easier for RAIL users and flag this as something we want to explore down the road in terms of how we deal with all of the RAIL dependencies.

eacharles:
Ok, from K-T:

'My initial impression is that if the import usages are in so few places (one or two functions per module), then this is a quite reasonable thing to do.  I don't think having a complex dependency-checking __init__.py solves the same problem (or is necessarily desirable).'

eacharles:
Based on K-T's comment above, my vote would be to merge this PR.  @johannct are you ok with that?

-e

eacharles (approved):
Rocking the local imports, not rocking the boat.

PR #194, opened 2022-06-14
fix problem with RailStage help message

eacharles:
This was being caused by the __getattr__ method in DataStore.   Go figure.   I put in a hack to avoid this particular problem.


sschmidt23 (approved):
help now works on rail.core.RailStage and all others that I tried, so looks good.

PR #196, opened 2022-06-15
138 add mpi capabilities to the estimation modules

### LINKED ISSUE #138: Add MPI capabilities to the estimation modules
Use an iterator to allow the estimation modules to work in pallalel with chunks of data.

joselotl:
This adds some functions to core, and also uses flexzboost as an example. I think I can write easilly all the other estimators in parallel form.


sschmidt23 (approved):
Everything seems to be working, and notebooks have been updated to reflect all changes.  We should wait for @joezuntz to check whether things are working as intended with TXPipe before merging in, but I'll mark as approved.

PR #197, opened 2022-06-15
Issue/177/spec selection

### LINKED ISSUE #177: Add spectroscopic selections as a new degrader
Add spectroscopic selection functions as a new degrader.

yanzastro:
the spectroscopic selection functions are contained in `spectroscopic_selections.py` file; `rail/creation/degradation/success_rate_data/` contains success rate data which will be updated if new data are available; `examples/creation/example_SpecSelection_for_HSC.ipynb` is an example notebook about how to do spec selection on a catalog.


rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+import os
+
+
+class SpecSelection(Degrader):
+    """
+    The super class of spectroscopic selections.
+
+    Parameters
+    ----------
+    colname: string, the name of the spec survey
```
jlvdb: If I understand the code correctly, these parameters refer to the `selection` method. It would be better to describe the configuration parameters instead (`N_tot`, `downsample`, etc.).

Maybe also highlight that the `selection` method should be overwritten by the subclasses (could even define this as an abstract base class if you like)

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+        Selection functions
+        """
+
+    def downsampling_N_tot(self):
+        """
+        Method to randomly sample down the objects to a given
+        number of data objects.
+        """
+        N_tot = self.config["N_tot"]
+        N_selected = np.where(self.mask)[0].size
```
jlvdb: Just a suggestion: this is what `np.count_nonzero()` is perfect for

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+
+        self.add_data('output', data_selected)
+
+    def __repr__(self):
+        """
+        Define how the model is represented and printed.
+        """
+
+
+class SpecSelection_WiggleZ(SpecSelection):
```
jlvdb: This selection function is incomplete. WiggleZ uses some UV and WISE selections that we cannot replicate in DC2 easily, which is why the original implementation of this selection function uses an additional downsampling step that tries to replicate the redshift distribution of the WiggleZ data sample.
I suggest to either implement this downsampling as well or dropping this degrader entirely.

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+
+    name = 'specselection_gama'
+
+    def selection(self, data):
+        """
+        GAMA selection function
+        """
+
+        print("Applying the selection from GAMA survey...")
+        self.mask *= (data["mag_r_lsst"] < 17.7)
```
jlvdb: Shouldn't this cut be r<19.8 for GAMA?

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+
+    def selection(self, data):
+        """
+        BOSS selection function based on
+        http://www.sdss3.org/dr9/algorithms/boss_galaxy_ts.php
+        The selection has changed slightly compared to Dawson+13
+        """
+
+        print("Applying the selection from BOSS survey...")
+        mask = (np.abs(data["mag_g_lsst"]) < 99.0) & \
```
jlvdb: Maybe we should add an additional check here (and in the following) that removes NaN magnitudes as well? 99 is a substitute for a non-detection that might not be used by other parts of RAIL (except for BPZ maybe).

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+    def photometryCut(self, data):
+        """
+        Applying DEEP2 photometric cut based on Newman+13.
+        This modified selection gives the best match to the data n(z) with
+        its cut at z~0.75 and the B-R/R-I distribution (Newman+13, Fig. 12)
+        NOTE: We cannot apply the surface brightness cut and do not apply the
+              Gaussian weighted sampling near the original colour cuts.
+
+        """
+        mask = (
```
jlvdb, line 241: Do these cuts work for DC2? They were adapted from the literature DEEP2 selection to account for differences between the data and MICE2 SEDs.

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+        """
+        mask = (data["mag_i_lsst"] > 15.0) & (data["mag_i_lsst"] < 22.5)
+        # 15.0, 22.5
+        self.mask &= mask
+
+    def speczSuccess(self, data):
+        """
+        Spec-z success rate as function of redshift (x) and I_AB (y) read of
+        Figure 3 in Lilly+09 for zCOSMOS bright sample. Do a spline
+        interpolation of the 2D data and save it as pickle on the disk for
```
jlvdb: Should probably remove the pickeling statement since this feature has been removed from the code.

jlvdb: For that reason we can also remove the pickle file
`rail/creation/degradation/success_rate_data/zCOSMOS.cache`
that made it into the commits.

rail/creation/degradation/success_rate_data/_
```diff
@@ -0,0 +1 @@
+
```
jlvdb: I'm not sure where this file is generated. If it has no functionality I would recommend removing it from the repository.

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+        faster reloads
+        """
+        success_rate_dir = self.config["success_rate_dir"]
+        x = np.loadtxt(os.path.join(
+                success_rate_dir, "zCOSMOS_z_sampling.txt"))
+        y = np.loadtxt(os.path.join(
+                success_rate_dir, "zCOSMOS_I_sampling.txt"))
+
+        pixels_y = np.searchsorted(y, data["mag_i_lsst"])
```
jlvdb: If the data is smaller than the lowest bin edge in y, it will return the value 0. If the data is located within the first bin (between egde[0] and edge[1]), the value 1 is returned by `np.searchsorted`.
This behaviour will introduce a shift in the bin assignment below.

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+        pixels_x = np.searchsorted(x, data["redshift"])
+
+        rates = np.loadtxt(os.path.join(
+                success_rate_dir, "zCOSMOS_success.txt"))
+        ratio_list = []
+        for i, py in enumerate(pixels_y):
+            if (py >= rates.shape[0]) or \
```
jlvdb, line 412: I think the correct test should be to require: `py >= rates.shape[0] or py < 1`, the same applies for the x dimension.

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+               (pixels_x[i] >= rates.shape[1]):
+                rate = 0
+            else:
+                rate = rates[pixels_y[i]][pixels_x[i]]
```
jlvdb: To avoid the index shift from `np.searchsorted`, both indices (pixels_x/y) should be decremented by one.

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+                success_rate_dir, "zCOSMOS_z_sampling.txt"))
+        y = np.loadtxt(os.path.join(
+                success_rate_dir, "zCOSMOS_I_sampling.txt"))
+
+        pixels_y = np.searchsorted(y, data["mag_i_lsst"])
+        pixels_x = np.searchsorted(x, data["redshift"])
+
+        rates = np.loadtxt(os.path.join(
+                success_rate_dir, "zCOSMOS_success.txt"))
+        ratio_list = []
```
jlvdb: Could be more efficient to declare this as numpy array and use the loop index `i` to assign the rate values.

rail/creation/degradation/spectroscopic_selections.py
jlvdb: I saw you did that correctly in the HSC degrader, so you can use the same strategy here.

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,521 @@
...
+        rates = np.loadtxt(os.path.join(
+                success_rate_dir, "hsc_success.txt"))
+
+        pixels_y = np.searchsorted(y_edge, data["mag_g_lsst"]-data["mag_z_lsst"])
+        pixels_x = np.searchsorted(x_edge, data["mag_i_lsst"])
+
+        pixels_y = pixels_y - 1
+        pixels_x = pixels_x - 1
+
+        ratio_list = []
```
jlvdb: Numpy array as well here?

rail/creation/degradation/success_rate_data/hsc_success.txt
```diff
@@ -0,0 +1,200 @@
... 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00 0.000000000000000000e+00
```
jlvdb: Maybe add a header comment line starting with '#' to give some basic information about this data (as in the data files for zCOSMOS)?

rail/creation/degradation/success_rate_data/hsc_i_binedge.txt
```diff
@@ -0,0 +1,201 @@
+1.300000000000000000e+01
```
jlvdb: Maybe add a header comment line starting with '#' to give some basic information about this data (as in the data files for zCOSMOS)?

rail/creation/degradation/success_rate_data/hsc_gz_binedge.txt
```diff
@@ -0,0 +1,201 @@
+-2.000000000000000000e+00
```
jlvdb: Maybe add a header comment line starting with '#' to give some basic information about this data (as in the data files for zCOSMOS)?

jlvdb (changes requested):
Hi Ziang,
The general structure of the code is very good and should be ready to use, great work! Also I think your updated implementation of the zCOSMOS case (which used to be a bottle neck) should run much faster.

There are, however, a few issues that need to be resolved:
- The documentation of the meta class could be improved.
- Some minor issues in the GAMA and zCOSMOS selection function.
- There are some unused data files included in the pull request.

The main problem I see are the unit tests that you have implemented. It seems like you just run all your code to see if it raises any exceptions. What the unit tests should do is make sure that the output stay consistent at all time, i.e. code changes do not alter the computed values (unless that is intended). You could make some accidental changes to these selection functions and the code would always pass the unit test instead of notifying the developer that the results have changed fundamentally, which they should not.

rail/creation/degradation/spectroscopic_selections.py
sschmidt23: Note that in many of the estimator codes we have `nondetect_val` as a stage parameter, e.g.
https://github.com/LSSTDESC/RAIL/blob/e3e6ed6a0dd0579e0a676d546bc7026dba3b547b/rail/estimation/algos/flexzboost.py#L86
That way we can have the code handle 99, np.nan, -99, etc...
Probably a good idea to do something similar here.

sschmidt23:
@aimalz will probably have opinions on the locations of the txt files and keeping code separate from parameter/initialization files.  I think we've been keeping files like e.g. `hsc_i_binedge.txt` (but including all of the txt files that just contain lists of numbers) in examples/creation/data rather than in with the code in the rail/creation/degradation path.  Separating them does keep the code directory less cluttered.  Alex can say whether or not he agrees here.

rail/creation/degradation/spectroscopic_selections.py
yanzastro: I have added this parameter in config.

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -61,16 +63,36 @@ def validate_colnames(self, data):
...
     def selection(self, data):
         """
-        Selection functions
-        """
-
+        Selection functions. This should be overwritten by the subclasses 
+        corresponding to different spec selections.
+        """
+
+    def invalid_cut(self, data):
```
sschmidt23, line 83: I don't think that we want this, there are legitimate reasons for non-detections for perfectly valid objects e.g. Ly Break galaxies.  Instead of removing the galaxies, I think we want some way of replacing the "non-detection" values with another value.  For example, in FZBoost we replace the non-detection with the 1 sigma flux limit (which is a user-definable parameter but have default values), and the error with 1.0. https://github.com/LSSTDESC/RAIL/blob/e3e6ed6a0dd0579e0a676d546bc7026dba3b547b/rail/estimation/algos/flexzboost.py#L55
Outright removing all galaxies not detected in all six bands could introduce biases into the sample.

sschmidt23:
Do you really want to check in TEMPZFILE.out, examples/creation/output.pq, examples/estimation/data/FILTER/DC2LSST_z_gaussian_coefficients.txt, and a few other files?  I think several of these are incidental files created when the tests and/or notebooks are run but that should not actually be included in the repo.

aimalz:
If any filter conversions are added to the `core` code, then this also might close LSSTDESC/rail_astro_tools#1.

yanzastro:
> If any filter conversions are added to the `core` code, then this also might close LSSTDESC/rail_astro_tools#1.

I'm trying to convert DEEP2 color to lsst color. Do you mean that it has already been implemented in the `core` code?

aimalz:
@yanzastro No, sorry for any confusion! I was saying that you can close two issues with just this one PR if you implement the conversion function generically.

sschmidt23:
I do think that we probably want the conversions to be separate from the spec_selection, as I think it could be generally useful for other applications beyond just these specific selections in future analyses.  Maybe a separate standalone degrader?

yanzastro:
@sschmidt23 @aimalz I added a config parameter `colnames` to tell the selection degrader which columns are necessary for the selection. It depends on which spectroscopic survey we want to use and can be an added column from a conversion degrader.

I also added a config parameter `nondetect_val` to specify the non-detection value as Sam suggested.

sschmidt23:
Ok, great.  When you think that you are finished go ahead and re-request a review from me and I will take a look at all of the changes.

sschmidt23 (changes requested):
Several files seem to just be linearly spaced arrays that should just be constructed with np.linspace or np.arange, the files should be deleted and replaced with the corresponding linspace/arange, no need to store and load them.  I think these are:
`hsc_i_binedge.txt`
`hsc_gz_binedge.txt`
`zCOSMOS_I_sampling.txt`
`zCOSMOS_z_sampling.txt`

It looks like spectroscopic_selections.py contains several instances of random number usage from numpy, but no seed is initialized.  For reproducibility I would say you should add a seed as a config parameter to those degraders so that we can reproduce exact samples when necessary.

I'm still a little uncomfortable with the fact that all of the filters are hardcoded with generic names of `g`, `r`, `i`, etc... when each survey being modeled has slightly different filters that correspond to different photometric systems, so `r` in HSC does not mean the same things as `r` in DEEP2.  If we're going to have hardcoded filter names and numbers for specific surveys, then maybe we should barcode filter names like `HSC_I` rather than `i`, and thus force the user to transform the input or else an error is raised.  I think this would essentially break the degraders until we implement said transforms, but then again, we aren't actually reproducing these surveys until we have those filter transforms anyway.  Maybe we approve this PR and merge this in now, but create an issue saying that we should update these degraders to have the filter names changed once the filter/photometric transforms are available.

yanzastro:
Hi @sschmidt23 , thank you for reviewing the PR. I'll take care of the txt files. For the column names, note that `colnames` parameter is a dictionary like `{'u': 'column_name_of_u_band', 'r': 'column_name_of_r_band', ...}`. The keys of the dict are generic, but their values (the column names in the data table) depend on surveys. For example, if we have a conversion degrader in the future which generates a column called `deep2_r` for `r` band of DEEP2, then we can set `colnames={'r', 'deep2_r',...}` to tell the selection degrader to take the `deep2_r` column as the `r` band of DEEP2 to make the selection.

sschmidt23:
Ah, ok, my mind glossed right over the self.config in front of the `i`, `g` when reviewing, sorry about that.

sschmidt23:
Oh, also, in the `example_SpecSelection_for_zCOSMOS.ipynb` I think you need to add a `rail.creation.degradation` in front of the two imports of spectroscopic_selections, as the user will likely not have RAIL/rail/creation/degradation in their path (at least I didn't.  So, the 7th-ish cell should, I think, be:
```
import sys
from rail.creation.degradation import spectroscopic_selections
from importlib import reload
from rail.creation.degradation.spectroscopic_selections import SpecSelection_zCOSMOS
```

sschmidt23:
I think the other thing that we should really do before approving this pull request is see a plot of the color/magnitude/ redshift distributions of each of the surveys along side the results of running them on either pzflow-generated data or cosmoDC2 data to see that we are actually reproducing the distributions that we want.  I don't know that this necessarily has to go in the repo, but we should definitely do some verification tests of this sort for each of the degraders.

yanzastro:
Yes I thought about this, but the problem is that we don't have the conversion degrader yet. Directly using LSST filter might introduce bias in the selected sample. Shall we wait until we have such degrader? I can try to write one

sschmidt23:
From the tag-up this morning it sounds like the consensus is that we should approve this PR and then create a separate issue for the photomteric transformations that we can add in the near future.  

So I think the final thing before approving is switching out the setting of np.random.seed() in spec_selections.py to use 
`rng = np.random.default_rng(seed=self.config.seed)` instead.

sschmidt23 (approved):
Looks good, thanks, Ziang

rail/creation/degradation/spectroscopic_selections.py
```diff
@@ -0,0 +1,519 @@
...
+    name = 'specselection_VVDSf02'
+
+    def photometryCut(self, data):
+        """
+        Photometric cut of VVDS 2h-field based on LeFèvre+05.
+        NOTE: The oversight of 1.0 magnitudes on the bright end misses 0.2 %
+               of galaxies.
+        update the internal state
+        """
+        mask = (data[self.config.colnames['i']] > 18.5) & (data[self.config.colnames['i']] < 24.0)
```
jlvdb: Could you change the upper i-band limit to 17.5, which is the original value for VVDS?

jlvdb (approved):

rail/creation/degradation/spectroscopic_selections.py
yanzastro: I guess you mean lower limit of i-band from 18.5 to 17.5?

jlvdb: Yes, sorry for the confusion

PR #199, opened 2022-06-16
Dealing with two issues found while using TXPipe

joezuntz:
Hi all. 

I found a couple of issues when using the newest RAIL from TXPipe:
- the `name` attribute was missing from the BPZ_lite class, which is needed to run from a YAML file correctly
- FlowEngine sometimes produces objects with z < 0. This was causing a crash in the BPZ dN/dz fitting, so I added a cut on these.  Might also want to look at removing them on generation, but I wasn't sure how to do that.

I also added a couple of print-outs to give users feedback on progress, but if people don't like these I can remove them.

### DIFF
--- a/rail/creation/degradation/grid_selection.py
+++ b/rail/creation/degradation/grid_selection.py
@@ -124,6 +124,7 @@ class GridSelection(Degrader):
         flat_inds = [item for sublist in keep_inds for item in sublist]
         training_data = data_hsc_like.loc[flat_inds, :]
 
+        training_data = training_data[training_data['redshift'] > 0]
         # For the pessimistic choice, also remove galaxies with z > redshift_cut from the sample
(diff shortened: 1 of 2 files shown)


jfcrenshaw:
Ah thanks for pointing out that `FlowEngine` sometimes produces negative redshifts. This is probably because RAIL still uses an older version of PZFlow. I'll make sure this gets fixed in generation when I transition RAIL to the newer version.

joezuntz:
I might just have the wrong version of PZFlow installed locally - should I just try upgrading?

eacharles (approved):
Thanks Joe

jfcrenshaw:
> I might just have the wrong version of PZFlow installed locally - should I just try upgrading?

This probably won't fix it since RAIL is currently pinned to an older version of PZFlow. I need to update this. But the fix you put in is fine for now.

PR #201, opened 2022-06-17
Issue/189/renaming in creation (adding Modeler class)

### LINKED ISSUE #189: Refactoring stages for creation
@jfcrenshaw and I were discussing how to unify the degraders in the context of creators (i.e. can we recycle the existing infrastructure to generalize the notion of degradation in terms of probability distributions?) and noticed that the default stages for creation still have names like `Engine` where it now means `Sampler` and that there isn't a way to train a new model, which is what `Engine` was originally referring to. (Also `Evaluator` appears in creation but means something else within RAIL already, so we can fix that along the way.) We're going to hack through making some consistency changes in creation on Monday.

aimalz:
This is mostly renaming "engine" because its meaning got lost in the ceci conversion, but there are still stages for "model --> catalog" and "model --> probabilities" as well as "catalog --> catalog" (degradation). It will remain a draft PR, however, until @jfcrenshaw and I implement the third necessary `RailStage` under `rail.creation`, for "config --> model," for which we currently have no example because we always started from having a pre-trained flow, but which is important for creators currently under development.


sschmidt23 (changes requested):
Can one of you summarize the major changes that I should look at? As this PR is rather old the git diff has a bunch of cruft from other changes mixed in, and I don't really want to dig through a lot of those bits that are actually from other PRs.

Several of the tests are failing, that should be fixed before I really review, I believe this is related to the Slack discussion on aliases for model files here: https://lsstc.slack.com/archives/CQGKM0WKD/p1661210015317229

Also, just looking through the list of files changed, I see goldenspike\ copy.ipynb, is that meant to actually be checked in here?  My guess is no.

jfcrenshaw:
Yeah I still need to fix the tests with aliases. I will try that out this week.

Once I have that done, I will make a list of things you might want to look at.

jfcrenshaw:
@sschmidt23 the things to look at are

- the `Modeler` class in `creation/engine.py`
- the `FlowModeler` class in `creation/engines/flowEngine.py`
- the example using `FlowModeler` at the top of goldenspike
- the example yaml file for FlowModeler at `examples/creation/configs/flowModeler.yaml`

@aimalz please speak up if there is anything else to look at

sschmidt23 (approved):
Looks good, approved

PR #203, opened 2022-06-21
Update grid_selection.py

imoskowitz:
Added in two new parameters: color_based_redshift to specify if we make a color-based redshift cut; percentile_cut: if doing a color-based redshift cut, the percentile to use for spec-z galaxies to determine the maximum redshift in a pixel.

Updated default file for ratios to include the file with spec-z galaxy information.

Made the pessimistic redshift cut before doing the random selection.


sschmidt23 (approved):
This now looks good, I added a reference to the HSC 2nd data release to both the grid_selection.py file, and created a README.md in examples/creation/data that lists the reference.  
Approving, and will merge momentarily.

PR #204, opened 2022-06-22
Speed up Inform_BPZ_lite stage

joezuntz:
This adds a simple parallelization to the slow part of the `Inform_BPZ_lite` stage, which is the fitting of the n(z) parameters. The function parallelizes very simply, since the log-like is just a sum over values from each spec-z galaxy.

Every ceci stage automatically has `rank` and `size` attributes which come from MPI.


joezuntz:
Don't review yet not ready!

joezuntz:
This is ready for review now. The code coverage seems to be a false alarm - it's not a file related to this PR.

eacharles (commented):
Generally looks fine to me.   I'm not sure I understand why some of the changes got included in the diff.   Perhaps the way we did the merging?

sschmidt23 (approved):
Looks good, code is a factor of >20 faster now

PR #205, opened 2022-06-22
Fix PZFlow Seed

joezuntz:
Currently the FlowEngine stage does not correctly pass the `seed` option to PZFlow, because it is set as a positional argument in the wrong position. Previously all runs used the same `None` seed.

It looks like a seed of `None` in PZFlow was intended to mean "generate a seed", but that doesn't seem to have been happening on my machine at least because I found this bug because two runs with different seeds generated the same sample.

By chance the argument that was previously being set (`conditions`) was being ignored because another arg was not supplied.


eacharles (approved):
Yoopie, we have random numbers!!

PR #207, opened 2022-06-23
issue/206/update-pzflow

### LINKED ISSUE #206: Move to PZFlow v3
RAIL currently uses PZFlow version 2. I need to make it compatible with version 3. Working on this now.

jfcrenshaw:
Updating RAIL to be compatible with PZFlow v3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import find_namespace_packages, setup
 # basic dependencies for all RAIL modules
 install_requires = [
     "h5py",
-    "numpy",
+    "numpy<1.23",
     "pandas>=1.1",
     "tables-io>=0.7.5",
     "ceci",
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-from setuptools import setup, find_namespace_packages
+from setuptools import find_namespace_packages, setup
 
 # basic dependencies for all RAIL modules
 install_requires = [
@@ -11,7 +11,7 @@ install_requires = [
 ]
 
 # dependencies for the Creation module
-creation_extras = ["pzflow>=2.0.7,<3.0.0"]
+creation_extras = ["pzflow"]
 
 # dependencies required for all estimators in the Estimation module
 estimation_extras = [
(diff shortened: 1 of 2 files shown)


jfcrenshaw:
The tests pass on my machine. I will investigate.

sschmidt23:
I just pushed a trivial linting change in a separate PR branch and am also getting test failures, maybe a push to tables_io/qp/etc... broke something, or some package like scipy updating caused a change, but it's not just your code.

sschmidt23:
oh wait, the two tests that fail are the two pzflow tests. Hmm, it looks like the creation_extras still has pzflow pinned to the specific older version in Irene's PR, so I'm not sure what is causing the problem there...

sschmidt23:
oh wait, the two that failed are the two pzflow tests.  Oh, and Joe Z. just pushed a fix to the fact that we weren't passing the random seed properly:
https://github.com/LSSTDESC/RAIL/commit/7ceaa7d308f12f58781597a534393b55f21ca8aa
So, I'm guessing we now just have different randoms, and thus the zb_expected in the test are now wrong because they have a new seed.  So, we probably just need to update those values in the test to what they should be in the new test.

jfcrenshaw:
Ah okay, I'l look at this tomorrow afternoon

jfcrenshaw:
Still failing on the github action. It's a weird jax error. ~I think I might be able to fix this by rewriting `PZFlowPDF`. I'll do this next week, including some of the upgrades in issue LSSTDESC/rail_pzflow#4~ 

`numpy 1.23` has some bugs (e.g. [here](https://github.com/numpy/numpy/issues/21838) and [here](https://github.com/numpy/numpy/issues/21840)) and made some changes that caused jax to break (e.g. [here](https://github.com/google/jax/pull/11230), [here](https://github.com/google/jax/issues/11221), [here](https://github.com/google/jax/issues/11253)). It looks like a fix has been implemented in jax, but these fixes haven't been published on PyPI yet. In addition, it looks like numpy hasn't finished fixing their bugs yet. For now, I have pinned `numpy<1.23`, which solves our problems. I will open an issue reminding us to remove this pin in a week or two.

sschmidt23 (approved):
Hopefully at some point the jax people will fix the numpy errors and we can drop the numpy<1.23 requirement, but this fixes things for now.

Issue #209, opened 2022-06-27
Add parallelization to the summarizers


sschmidt23:
The existing summarizer codes currently assume that you will load all of the data at once and use only one process to create the summarized ensemble.  Three of the four existing summarizers (NZDir, NaiveStack, PointEstimateHist) have a major aspect that easily leads itself to be parallelizable (not so sure on varInference and the iterative determination of the distribution, that seems a bit trickier):
PointEstimateHist and NZDir are simply adding points to a histogram
NaiveStack is just adding to a gridded parameterization the sum of individual PDFs evaluated on said grid. 

So, a method that takes each chunk and does the histogram/grid sum combined with a scheme to combine each of the chunks into the final summed distribution should be fairly straightforward.

One complication: the branch that I plan to turn into a PR to be merged soon, [issue/173/summ_errors](https://github.com/LSSTDESC/RAIL/tree/issue/173/summ_errors), changes the format of all four summarizers to return N samples as an ensemble rather than a single PDF, where the samples are based on bootstraps.  Computing bootstraps in chunks is a little more complicated, we should probably look up the "bag of little bootstraps" technique, and try to figure out the optimal way to return the N samples with proper bootstrap statistics.  For PointEstimateHist and NaiveStack, the bootstrap is done on the large number of PDFs, and the samples are going to be quite similar, NZDir is the more realistic case with samples determined by the smaller spectroscopic sample, which will lead to more reasonable sample uncertainties via the bootstraps.  (i.e. I think NZDir is the summarizer we should actually focus on in terms of implementing the bootstraps properly).

joselotl:
Hi @sschmidt23 
I'm starting to work in this issue. I saw that NZDir has the bootstrap implemented over the training dataset. But I think that doing the "bag of little bootstraps" technique sampling over the training dataset will be unrealistic, because each node will need to use the full photometric data set.
Maybe doing the sampling over the photometric data will be better. How large will be the final training dataset? Is it feaseble to send to each node the full training dataset?

sschmidt23:
My guess is that the training set will be big but not nearly as big as the photometric datasets, probably on order of 100,000 - 1,000,000 galaxies or so (there are larger samples of bright galaxies, e.g. SDSS, but there's no reason to use all of them when we only have 10,000-100,000 faint galaxy spec-z's where the bulk of our data actually live in parameter space, I think).  So, I think it still might be feasible to try bootstrapping the spec-z sample.

[...]

PR #210, opened 2022-06-27
Convert Summarizer codes to return N samples rather than a single PDF distribution

### LINKED ISSUE #173: Bootstrap errors on the redshift distributions
As discussed in today's RAIL tag up, it would be useful for us to provide errors on the n(z) that we obtain from the stacking and DIR methods as part of the STAR challenge. This could be in the form of a large number of bootstrap realizations of the n(z) consistent with past approaches in the literature (e.g. [KiDS-450](https://arxiv.org/pdf/1606.05338.pdf)). 

@eacharles @sschmidt23 @aimalz

sschmidt23:
This PR includes changes to NZDir, VarInferenceStack, NaiveStack, and PointEstimateHist that have them return N bootstrap samples rather than a single PDF for the sample (along with a second file that contains the fiducial single PDF, along with a 1 sigma error estimate).  For VarInf, Naive, and PointEst, the bootstraps are done on the photometric/input sample, as there is no spec-z sample to bootstrap, for large samples this means that the bootstrap error will be tiny, but that's the best we can do.  NZDir does a "proper" bootstrap of the smaller spectroscopic sample. 

Currently, the fiducial PDF contains a +/- 1 sigma error at **each point** in the histogram/grid, rather than something like a single number error on the mean redshift.  We could swap out the error estimate to be the more simple single number if that is desired.  The fiducial estimate seems almost superfluous, since it's trivially calculable from the N bootstraps, but it was requested, so that second output file remains.

Also, this implementation includes no parallelization, and requires all of the data to be read in at once, I've created  a separate issue, #209, mentioning this desired functionality, which we can add in the future.


sschmidt23:
Oh, and we may want to delete or clean up the example ipynb, I didn't know how useful that was in the long run, I mainly made it to be instructive for some chats we were having for the STAR challenge.

rail/estimation/algos/NZDir.py
```diff
@@ -162,7 +167,40 @@ def run(self):
...
+        nsamp = self.config.nsamples
+        bootstrap_indeces = np.random.randint(ngal,
+                                              size=ngal * nsamp).reshape([nsamp, ngal])
+        hist_vals = np.empty((0, self.config.nzbins))
+        for i in range(nsamp):
+            uniq, cnts = np.unique(bootstrap_indeces[i], return_counts=True)
+            zarr = np.array([])
+            tmpweight = np.array([])
+            for un, ct in zip(uniq, cnts):
+                zarr = np.concatenate((zarr, np.repeat(self.szvec[un], ct)), axis=None)
```
joezuntz: When you do np.concatenate like this each time it repeatedly allocates, copies, and then deallocates memory.

In this case if I'm reading it right you know the total size (`cnts.sum()`) so it's better to pre-allocate the full array.

Looks like this applies to the outer loop also?

rail/estimation/algos/naiveStack.py
```diff
@@ -15,16 +17,40 @@ class NaiveStack(PZSummarizer):
...
+        for i in range(self.config.nsamples):
+            uniq, cnts = np.unique(bootstrap_draws[i], return_counts=True)
+            single_bval = np.zeros(len(self.zgrid))
+            for j, un in enumerate(uniq):
+                single_bval += pdf_vals[un, :] * cnts[j]
+            bvals = np.vstack([bvals, single_bval])
+        sample_ens = qp.Ensemble(qp.interp, data=dict(xvals=self.zgrid, yvals=bvals))
+
+        # compute 1 sigma errors
+        pdf_vals = sample_ens.pdf(self.zgrid)
```
joezuntz: Is it worth moving this part to a separate function or superclass method since it's being repeated many times?  Also true for the bootstrap index choice code above.

joezuntz:
I've just been testing this, and find that somehow these changes make NZDir not work as expected - on this branch the n(z) for each bin always comes out approximately the same, and basically just looks like the overall prior (without tomographic binning).

joezuntz:
Update: if I merge the main branch into this one it fixes the problem in the main n(z) single_nz file.  The 20 n(z) realizations still show the same thing.

joezuntz:
Another update: I found the issue here and pushed a fix. I think if you merge main into this then good to merge back.

I also found a speed improvement that makes it possible to get to thousands of realizations pretty quickly.

rail/estimation/algos/naiveStack.py
sschmidt23: After some discussion on Slack, I'm going to remove this error calculation from each of the code, as it's not actually all that useful, we will add a notebook in the near future with some example calculations of more useful uncertainty quantities.

rail/estimation/algos/NZDir.py
sschmidt23: Thanks for spotting this lazy coding from me, I fixed a similar implementation in pointEstimateHist as well

sschmidt23:
I want to make a few minor modifications to the demo notebook included in this PR, as well as the NZDir notebook, but after that I think this will be ready for another review.

sschmidt23:
I think this is now ready for re-review, I removed the not-so-useful uncertainty estimate from the fiducial file, made a few minor changes to the demo notebooks (mainly removing references to the now axed uncertainty estimate), and made a few other minor changes.  Things look good on my end, everything is working as expected.

rail/estimation/algos/NZDir.py
```diff
@@ -127,11 +132,13 @@ def open_model(self, **kwargs):
         self.sz_mag_data = self.model['sz_mag_data']
 
     def run(self):
+        np.random.seed(self.config.seed)
```
joezuntz: Using the global numpy random state occasionally causes problems for reproducibility running as part of a wider pipeline, because other code can modify it as well. They now recommend doing:
```
rng = np.random.seed(self.config.seed)
...
bootstrap_indices = rng.randint(ngal, size=ngal)
```

rail/estimation/algos/naiveStack.py
```diff
@@ -15,16 +17,31 @@ class NaiveStack(PZSummarizer):
...
         test_data = self.get_data('input')
-        self.zgrid = np.linspace(self.config.zmin, self.config.zmax, self.config.nzbins+1)
+        self.zgrid = np.linspace(self.config.zmin, self.config.zmax, self.config.nzbins + 1)
         pdf_vals = test_data.pdf(self.zgrid)
         yvals = np.expand_dims(np.sum(np.where(np.isfinite(pdf_vals), pdf_vals, 0.), axis=0), 0)
         qp_d = qp.Ensemble(qp.interp, data=dict(xvals=self.zgrid, yvals=yvals))
-        self.add_data('output', qp_d)
+
+        bootstrap_draws = np.random.randint(test_data.npdf, size=test_data.npdf * self.config.nsamples).reshape([self.config.nsamples, test_data.npdf])
+        bvals = np.empty((0, len(self.zgrid)))
```
joezuntz: You could probably speed this up in a similar same way to how I did with the NZDIR if you like.

rail/estimation/algos/naiveStack.py
```diff
@@ -15,16 +17,31 @@ class NaiveStack(PZSummarizer):
     config_options = PZSummarizer.config_options.copy()
     config_options.update(zmin=Param(float, 0.0, msg="The minimum redshift of the z grid"),
                           zmax=Param(float, 3.0, msg="The maximum redshift of the z grid"),
-                          nzbins=Param(int, 301, msg="The number of gridpoints in the z grid"))
+                          nzbins=Param(int, 301, msg="The number of gridpoints in the z grid"),
```
joezuntz, line 20: Is it worth having a seed on this stage too?

rail/estimation/algos/pointEstimateHist.py
```diff
@@ -16,16 +18,36 @@ class PointEstimateHist(PZSummarizer):
     config_options.update(zmin=Param(float, 0.0, msg="The minimum redshift of the z grid"),
                           zmax=Param(float, 3.0, msg="The maximum redshift of the z grid"),
                           nzbins=Param(int, 301, msg="The number of gridpoints in the z grid"),
-                          point_estimate=Param(str, 'zmode', msg="Which point estimate to use"))
+                          point_estimate=Param(str, 'zmode', msg="Which point estimate to use"),
```
joezuntz, line 22: Again is it worth having a seed here?

rail/estimation/algos/pointEstimateHist.py
```diff
@@ -16,16 +18,36 @@ class PointEstimateHist(PZSummarizer):
...
+        self.zgrid = np.linspace(self.config.zmin, self.config.zmax, self.config.nzbins + 1)
+        self.bincents = 0.5 * (self.zgrid[1:] + self.zgrid[:-1])
+        single_hist = np.histogram(test_data.ancil[self.config.point_estimate], bins=self.zgrid)[0]
         qp_d = qp.Ensemble(qp.hist,
-                           data=dict(bins=self.zgrid, pdfs=np.atleast_2d(hist_vals)))
-        self.add_data('output', qp_d)
+                           data=dict(bins=self.zgrid, pdfs=np.atleast_2d(single_hist)))
+        hist_vals = np.empty((nsamp, self.config.nzbins))
+        for i in range(nsamp):
+            bootstrap_indeces = np.random.randint(npdf, size=npdf)
```
joezuntz: I think this is probably much faster as `zarr = zb[bootstrap_indeces]`

rail/estimation/algos/varInference.py
```diff
@@ -45,21 +49,24 @@ def run(self):
...
-        log_pdf_vals = np.log(np.array(pdf_vals)+TEENY)
+        log_pdf_vals = np.log(np.array(pdf_vals) + TEENY)
         alpha_trace = np.ones(len(self.zgrid))
         init_trace = np.ones(len(self.zgrid))
 
         for _ in range(self.config.niter):
             dig = np.array([digamma(kk) - digamma(np.sum(alpha_trace)) for kk in alpha_trace])
             matrix_grid = np.exp(dig + log_pdf_vals)
-            gamma_matrix = np.array([kk/np.sum(kk) for kk in matrix_grid])
+            gamma_matrix = np.array([kk / np.sum(kk) for kk in matrix_grid])
```
joezuntz, line 61: Could you put a comment in here describing or linking to an explanation of the maths?

rail/estimation/algos/varInference.py
```diff
@@ -45,21 +49,24 @@ def run(self):
...
             nk = np.sum(gamma_matrix, axis=0)
             alpha_trace = nk + init_trace
 
+        # old way of just spitting out a single distribution
+        # qp_d = qp.Ensemble(qp.interp, data=dict(xvals=self.zgrid, yvals=alpha_trace))
+        # instead, sample and save the samples
+        sample_pz = dirichlet.rvs(alpha_trace, size=self.config.nsamples)
```
joezuntz: Similarly you can get a reproducible result here by doing
```
rng=np.random.default_rng(seed=self.config.seed)
sample_pz = dirichlet.rvs(alpha_trace, size=self.config.nsamples, random_state=rng)
```

joezuntz (approved):

joezuntz:
I've put some possible small improvement above, but broadly happy.  The notebooks all ran for me.

rail/estimation/algos/NZDir.py
sschmidt23: changed

rail/estimation/algos/naiveStack.py
sschmidt23: done

rail/estimation/algos/naiveStack.py
sschmidt23: added a seed

rail/estimation/algos/pointEstimateHist.py
sschmidt23: added

rail/estimation/algos/pointEstimateHist.py
sschmidt23: yep, my attempt at being clever just slowed things down, changed to the simpler way.

rail/estimation/algos/varInference.py
sschmidt23: Markus is writing up the paper now, we should have a reference to add in ~1 month, according to Markus, I'll create an issue to remind us to add the reference when the paper is finished.

PR #212, opened 2022-06-29
Restricted python to >= 3.8.

### LINKED ISSUE #211: Restrict python version to >= 3.8
`qp` requires 3.8 due to a scipy requirement to implement the mixmod parameterization.

I asked in #desc-co, and Mike Jarvis (endorsed by Michael Wood-Vasey) said we shouldn't worry about anything <3.8.

So I will bump the python requirement to >= 3.8, and update the testing matrix to [3.8, 3.9, 3.10]

jfcrenshaw:


jfcrenshaw:
@sschmidt23 said there was a "test-skip" for 3.7. Not sure if that lives in `qp` or in `rail`, but if it lives in `rail` we should remove it now.

sschmidt23:
It's actually a test on scipy version on [L229](https://github.com/LSSTDESC/RAIL/blob/20693573c2ea0e69f5deab958d8622d325b16ae6/tests/test_algos.py#L229) of test_algos that tests that scipy is above v 1.8.  As scipy 1.4-1.7 can still work with python3.8, I think we need to leave the test skip in; even if Github will grab the newer scipy versions for the automated tests via pip, some users could still have the older scipy on their machines.

eacharles (approved):

.github/workflows/main.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9, "3.10"]
```
sschmidt23, line 18: why is 3.10 in quotes here?

sschmidt23 (approved):
Looks good, as I noted, not sure why 3.10 is in quotes in the in the GH action, but that's minor, am approving.

PR #213, opened 2022-06-29
Issue/129/hyperbolic mags

jlvdb:
Added stages to the core module to compute hyperbolic magnitudes (Lupton et al. 1999). Includes:
- code
- example notebooks
- unittests


jlvdb:
See also issue #224 about the numerical precision in the unittests.

I had to add the external package `hyperbolic` to the `docs/requirements.txt` in order to be able to build the documentation.

eacharles (approved):
This looks really nice Jan Luca, thanks for your work.  
At some point int would be nice to have your hyperbolic package in pip or conda, I'd be happy to help you set that up.

jlvdb:
Thank you Eric. I was planning to do that eventually when the package is complete (there is some . I have published on pypi before, but I will come back to your offer when the time has come.

PR #216, opened 2022-07-05
fix errors, remove old sz write, switch rng form

### LINKED ISSUE #215: Fix a few warnings and extraneous bits in KNearNeigh
There are several annoying warnings that pop up with KNearNeighPDF because I was dumb and create a copy of a data frame.  There's also a write out of the spec-z that was there for debugging that should be removed.

sschmidt23:
Super simple PR to remove an unnecessary data frame copy that causes some annoying warnings; removal of writeout of the spec-z array that was left over from debugging, and switching the random number handling after Joe's comment about reproducibility in pipelines.


eacharles (approved):
Looks good, thanks!

PR #217, opened 2022-07-07
Issue/183/fspsinterface

torluca:
I'm creating this pull request to merge the FSPS interface into the main branch. I implemented some unit tests as well. By merging we could us the code coverage function of github to see whether other unit tests are necessary.


torluca:
The very first error I see concerns FSPS and its installation. I found that it works without any issue by running the following command:
git clone --recursive https://github.com/dfm/python-fsps.git
cd python-fsps
python -m pip install .

Is this implementable in setup.py or do we need to create a separate installation like we did for Delight?

tests/creation/test_sed_generator.py
```diff
@@ -0,0 +1,66 @@
...
+    [
+        ({"sfh_type": 3}, AssertionError),
+    ],
+)
+def test_FSPSSedGenerator_bad_tabulated_sfh_params(settings, error):
+    """Test bad tabulated sfh params that should raise Value and Type errors."""
+    with pytest.raises(error):
+        DS = RailStage.data_store
+        DS.__class__.allow_overwrite = True
+        dummy_io_data = tables_io.read('../data/test_fsps_sed_gen.fits')
```
sschmidt23: Two things: first, tests are usually run from the root RAIL directory, so paths should be relative to that not the location of the test file, so here this, I think, should be 'tests/data/test_fsps_sed_gen.fits'.  Second, `test_fsps_sed_gen.fits` is not included in the PR, so this test will fail.

tests/creation/test_sed_generator.py
```diff
@@ -0,0 +1,66 @@
...
+    [
+        ({"smooth_lsf": True}, AssertionError),
+    ],
+)
+def test_FSPSSedGenerator_bad_tabulated_lsf_params(settings, error):
+    """Test bad tabulated sfh params that should raise Value and Type errors."""
+    with pytest.raises(error):
+        DS = RailStage.data_store
+        DS.__class__.allow_overwrite = True
+        dummy_io_data = tables_io.read('../data/test_fsps_sed_gen.fits')
```
sschmidt23: Same as above comment, path should be set as if running pytest from the root RAIL directory

tests/creation/test_sed_generator.py
```diff
@@ -0,0 +1,66 @@
...
+        sed_generation_test = FSPSSedGenerator.make_stage(name='sed_generator_test', smooth_velocity=False,
+                                                          **settings)
+        sed_generation_test.add_data('input', dummy_io_data)
+        sed_generation_test.run(physical_units=True)
+
+
+def test_FSPSSedGenerator_output_table():
+    DS = RailStage.data_store
+    DS.__class__.allow_overwrite = True
+    dummy_io_data = tables_io.read('../data/test_fsps_sed_gen.fits')
```
sschmidt23: and here

sschmidt23:
It looks like from the documentation page https://dfm.io/python-fsps/current/installation/ that fsps is pip installable, so I think you can just add "fsps" to `creation_extras` in setup.py.  However, I see the install instructions also ask you to git clone Charlie Conroy's original repo (which is 2.3 GB!) and set an env var pointing to the location of the cloned repo, I'm not sure of the best way to do that in a setup.py, hopefully someone else has an idea?

eacharles:
Well, one interesting point is that the ceci design lets us grab stages from plug-in modules.  I'm thinking that between delight and Charlie Conroy's 2.3 GB of data, we should start thinking about breaking RAIL up into a package with core modules that are relatively easy to install and then have other package for the bits that are a mess to install.

tests/creation/test_sed_generator.py
torluca: Thanks Sam, I forgot to add the test file. Now it is present and the path to it is absolute.

tests/creation/test_sed_generator.py
torluca: Same here

tests/creation/test_sed_generator.py
torluca: Same

torluca:
Following RAIL restructuring, the interface to FSPS has been moved to the rail_fsps package. 
It can be found at https://github.com/LSSTDESC/rail_fsps.
The interface is working and a demo notebook has been added.

PR #218, opened 2022-07-07
add nthread to regression params

sschmidt23:
draft just to see if adding nthread param to regression params makes the GH action run faster.  It doesn't seem to affect cpu time, but does sometimes affect wall time in my tests.  And, I missed that the default value is `-1` not `1`, which sets it to try to use all available processors.  Setting default to 1 might speed things up.


sschmidt23:
Tinkering with the nthread value did not fix the long runtime problem, tlking to Biprateep, it sounds like xgboost has actually redefined some keywords.  It also sounds like the CMU group is going to try creating a new fork of FlexCode that will be better maintained, so I'm closing this PR and will watch for the updated FlexCode to become available at some point.

PR #221, opened 2022-07-12
Save zmean as well as zmode in BPZ_lite

joezuntz:
This adds zmean to the ancil data in BPZ_lite, so it is saved as a new column.


eacharles (approved):
🕺

PR #226, opened 2022-07-25
Add simple SOM summarizer via Issue/225/simplesom

sschmidt23:
This PR adds a SOM-based summarizer similar in many ways to the NZDir summarizer.  The difference here is that, rather than searching for neighbors around a set of spec-z objects and creating a weighted sum as the estimate, this code trains a SOM based on a large input set of data, then finds the best SOM cell for the spec-z and photometric sets, and creates a redshift summary estimate of the histogram (with optional weights) of spec-z objects in each cell weighted by the number of photometric galaxies (with optional weights) in that same cell and summing over all SOM cells.  The SOM code used here is MiniSOM, a lightweight, minimal SOM package that depends only on numpy and can be pip installed with a simple `pip install minisom`, which is a nice feature.

One nice feature is that you can identify SOM cells that contain photometric data but no spec-z members, and are thus uncovered and not properly calibrated by the summarizer (NZDir as coded lacks this and gladly chugs along ignoring that potential bias).  I'm marking this as a draft PR because currently I just spit out an array of the (raveled) cell indices of these uncovered cells, not sure if we actually want to add something like the IDs, line numbers, etc... for those objects, as presumably we want to identify and remove them from the input dataset.

The `Inform_SimpleSOMSummarizer` that builds a SOM could be generally useful in producing a SOM.  For example, you can do the first half of what SimpleSOMSummarizer does, identify uncovered cells, and then use that as a standalone selection step before running NZDir or other spec-z based summarizer.  You could also derive "SOM cell weights" to try to correct a spec-z set to have the same proportional fractions as a photometric set before using as input to an estimator code.


johannct:
I just committed small typos. som_sigma argument is not explained.

johannct:
"This demo used a relatively small dataset (~150,000 DC2 galaxies from one healpix pixel) to train the SOM, and even smaller photometric and spectroscopic datasets of 10,000 and 20,000 galaxies." Looks like the sentence is not complete?

sschmidt23:
There are a ton of commits from today to this PR where I was messing around trying to get readthedocs to compile the docs properly, the net result of all of those changes is essentially adding a .readthedocs.yaml to the top level directory to tell it to use python3.9 rather than (current default for readthedocs) python3.7, as our new scipy requirement no longer supports python3.7.

yanzastro:
I have finished the review and didn't find any major points in the code itself. It would be nice if we could easily implement other SOM codes (from DES or KiDS) to replace minisom (or even make it possible for the user to choose which SOM code to use). I see the current code is already pretty adaptable. Well done!

Some minor comments:

In `simpleSOM.py`:

1. `m_dim=Param(int, 31, msg="number of cells in SOM y dimension"),
n_dim=Param(int, 31, msg="number of cells in SOM x dimension"),`

Probably more informative for users to add some instruction in msg about how to choose the dimensions (typically sqrt(5*sqrt(number_of_input_vectors)))

2. Can we be more flexible with the input vectors? (like n magnitudes + m colors)

3. in the `run` function, can you add flagship to the sources in the photometric data that are mapped onto bad pixels? These sources should be removed for subsequent n(z) calibration.

In `SimpleSOM_demo.ipynb`

1. It's clearer to compare n(z) from SOM with true n(z) to draw them in one figure using `plt.plot` rather than histograms. You can even overplot the bootstrap samples with thinner lines on the same plot, something like:

```
fig, axs = plt.subplots(1,1, figsize=(12,8))
hist_true, zedge = np.histogram(bright_data.data['photometry']['redshift'], bins=np.linspace(0,3,101), density=True)
axs.plot((zedge[1:]+zedge[:-1])/2, hist_true, label='true N(z)')

for i in range(boot_ens.shape[0]):
    if i == 0:
        axs.plot(np.linspace(0,3,101), np.squeeze(boot_ens[i].pdf(np.linspace(0,3,101))), lw=1, color='red', alpha=0.3, label='SOM bootstrap samples')
    else:
        axs.plot(np.linspace(0,3,101), np.squeeze(boot_ens[i].pdf(np.linspace(0,3,101))), lw=1, color='red', alpha=0.3)
    
axs.plot(np.linspace(0,3,101), np.squeeze(bright_fid_ens.pdf(np.linspace(0,3,101))), color='k', label='SOM mean N(z)')
axs.legend(fontsize=15)
axs.set_xlabel('redshift', fontsize=15)
axs.set_ylabel('N(z)', fontsize=15)
```

2. maybe also add some cells to quantitatively evaluate the bias and errors of the SOM mean n(z)

sschmidt23:
For the various points:

1) what is the source of the `sqrt(5*sqrt(number_of_input_vectors))`, is there a paper reference for this?  I've seen similar formulae before but any explanation seemed very ad-hoc and somewhat specific to the input data (i.e. different data sets will have different inherent embedded dimensionality and might not conform to such a formula, I'd think).

2) Do we really want to be more flexible?  A SOM is a dimensionality reduction technique, and the fundamental amount of information is the number of input features, e.g. 6 magnitudes or 1 magnitude and 5 colors.  Using different combinations of magnitudes and colors does not change the actual information contained.  Using either 1 magnitude + N-1 colors or just the N-1 colors I think covers the same information content as n mags + m colors (I included "only colors" because C3R2 ignores overall magnitude information and only deals with color, thus throwing away some information).  Why do you think that n mags + m colors is a good option to have, just for increased flexibility? 

3) I was debating what to do about the cells flagged as bad.  Right now I just write out the (raveled) cell-IDs of cells that have photometric data but no spec-z objects, but I could instead output either the objID/galaxy_id for all objects not covered, or the full data vector of said objects.  I went with just the cell-IDs to be more compact in storage, but I will post to #desc-pz-rail to ask others what they think will be useful and we can go with what people want.

For the simplsom demo notebook, I want to stick to plotting histograms, as that is the actual native format for the estimator, there is a bug in qp (as far as I can tell, maybe I'm doing something wrong) does not adjust alpha/face color of the histograms, I will create an issue for this, which could make plotting more transparent (literally).

Adding some quantitative cells to the notebook is a good idea, I'll do that now.

sschmidt23:
@yanzastro I added a very simple plot of the bootstrap means and std devs vs the truth at the end of the demo notebook, what other metrics/plots do you think I should add here?

yanzastro:
@sschmidt23 Thanks for the reply!
For point 1, I can only find the reference here http://www.giscience2010.org/pdfs/paper_230.pdf saying that this choice can ensure enough cells while avoiding too many bad cells. I'm not sure about the exact mathematical reason behind it, but I'm not suggesting fixing the SOM size to be this number but add it as a recommendation for the user (I usually got confused when using other people's code on how to choose these setup so I think a bit of reference would make the code more user-friendly).

2. In the paper http://arxiv.org/abs/1909.09632 (SOM application for KiDS), the authors actually used 39 colors(all the combinations from 9 bands)+1magnitudes to train the fiducial SOM. In Appendix A.1.3, they discussed other choices of color-magnitude combinations. The idea behind this is that different options might improve or add too-much redundancy for the SOM. Adding flexibility on input allows us to do a similar test for the RAIL calibration.

3. Also in http://arxiv.org/abs/1909.09632, the authors remove the photometric sources mapped into a cell that has no spectroscopic counterpart and get a "gold sample" with guaranteed representation, so I suggested flagging out these sources (outputing either the objID/galaxy_id for all objects not covered is a good idea). I would also like to hear what other people think.

The new plots look nice, but I guess the std here is the std of each bootstrap sample? If this is the case, then it shows the width of n(z) in each sample, right? What I suggested is to calculate the std of the mean redshift from the 25 bootstrap ensembles like `np.std(boot_ens.mean().flatten())` (so we can get the information in Table.3 from the SOM paper I mentioned in point 2)

sschmidt23:
Ok, thanks @yanzastro  for the references, I will add a mention of the 5 *sqrt(number of rows x number of columns of data) and a reference in the documentation in the notebook.  We can also add a link to the KiDS paper for further reading. 

We can certainly add some flexibility in magnitudes and colors, while it seems weird to me to artificially increase dimensionality in a dimensionality reduction algorithm, SOMs are a bit weird, and I guess it could give better results.  I'll think about the most straightforward way to implement this.

Joe Zuntz mentioned that having the cell ID of every galaxy would make exploration and removal of "non-covered" objects easier to explore, I will add an (optional) output that writes out the galaxy ID and winner/cell ID for every object.

I will also add a Gaussian showing the std dev of the bootstraps to the example notebook.  I'll message you when I get these changes implemented (or if you review this PR and select "request changes" then I'll just re-request your review when the changes are ready.

yanzastro:
@sschmidt23 sounds good! add the cell ID is a great idea!

do I just click "request changes" after I add comments on associated lines in the reviewing mode? (sorry this is my first time reviewing a PR)

rail/estimation/algos/simpleSOM.py
```diff
@@ -0,0 +1,277 @@
...
+def_cols = [f"mag_{band}_lsst" for band in def_bands]
+def_maglims = dict(mag_u_lsst=27.79,
+                   mag_g_lsst=29.04,
+                   mag_r_lsst=29.06,
+                   mag_i_lsst=28.62,
+                   mag_z_lsst=27.98,
+                   mag_y_lsst=27.05)
+
+
+def _computemagcolordata(data, ref_column_name, column_names, justcolors):
```
yanzastro: Please add flexibility on the number of input colors and magnitudes for SOM

examples/estimation/SimpleSOM_demo.ipynb
```diff
@@ -0,0 +1,684 @@
...
+    "ax1.set_xlabel('redshift', fontsize=12)\n",
+    "ax1.set_title('mean and std for bright sample', fontsize=12);"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "64791557-7832-4a39-9a45-c3555fe15d28",
+   "metadata": {},
+   "source": [
+    "We see that the mean (red) and std dev (blue) estimates are quite biased compared to the truth in both cases, this is not unexpected for a simple true redshift cut and small samples sizes used in this demo.  The std dev estimate is much closer to the truth for the bright sample due to the elimination of the false secondary redshift peak.  However, we are still well above tolerances for any cosmological analysis, actual tomographic samples will need both tight color selections and larger training sets than those used in this demo in order to properly calibrate our redshift distributions."
```
yanzastro, line 672: add std of the mean redshift calculated from the bootstrap samples

rail/estimation/algos/simpleSOM.py
```diff
@@ -0,0 +1,277 @@
...
+        else:  # pragma: no cover
+            raise KeyError(f"spectroscopic weight column {self.config.spec_weightcol} not present in data!")
+
+        # find the best cells for the photometric and spectrosopic datasets
+        phot_colors = _computemagcolordata(test_data, self.ref_column_name,
+                                           self.usecols, self.use_only_colors)
+        spec_colors = _computemagcolordata(spec_data, self.ref_column_name,
+                                           self.usecols, self.use_only_colors)
+
+        phot_som_coords = np.array([self.som.winner(x) for x in phot_colors]).T
```
yanzastro, line 290: add cell ID of each source/galaxy into the data catalog

yanzastro (changes requested):
please check and revise the commented lines in the code. Very nice work!

examples/estimation/SimpleSOM_demo.ipynb
sschmidt23: Done, calculated np.std and put a Gaussian on the plot showing the spread visually

rail/estimation/algos/simpleSOM.py
sschmidt23: I added another output file that you specify with 'cellid_output', it checks for an ID column, if present it writes that, and always writes out the X, Y 2D cell coords and the raveled 1D cell coords for each galaxy from the photometric sample.

rail/estimation/algos/simpleSOM.py
sschmidt23: I switched out the `use_only_colors` boolean flag for `column_usage` which you can set to `colors`, `magandcolors`, or `columns`.  If you choose columns then the SOM will be constructed from all of the columns specified in the `usecolumns` list.  That way, you can put in whatever combination of inputs you want, as long as you construct the same set for the training, photometry, and spec-z samples.  I tested this with 6 mags and 5 colors and it worked as expected, you just need to remember to also specify a `nondetect_val` dictionary with an entry for each of the columns in `usecols` (I put some documentation on this in both the simpleSOM.py and the demo notebook).

sschmidt23:
I think that I have addressed all of the comments, take a look at the newer changes and see if this is satisfactory, or if we need to make a few more additions.

yanzastro (approved):

PR #227, opened 2022-07-29
Issue/133/extended magerr

### LINKED ISSUE #131: Add a new degrader to model magnitude error for extended sources
The current `LSSTErrorModel` only deals with point sources. We'll add a new module to deal with extended sources with finite aperture sizes.

The aperture sizes are calculated with two models:
1. the model from https://arxiv.org/abs/1902.11265, that is, to assume the magnitude to be measured as Kron-like elliptical aperture magnitude;
2. the Gaussian Aperture and PSF model from https://arxiv.org/abs/1902.11265, that is, to assume the magnitude to be measured as GAaP magnitude.

This is part of the GCCL in-kind contribution.

yanzastro:


jfcrenshaw:
@yanzastro great work! I cleaned up some of the documentation and fixed a few tests so that their outputs are deleted at the end of the test.

Two questions:
1. You committed a lot of files in the `examples/estimation` directory that look like they're outputs from the example? i.e. `NZDir_NZ.hdf5`, `NZDir_samples.hdf5`, `NaiveStack_NZ.hdf5`, `Naive_samples.hdf5`, `demo_knn.pkl`, `nzdir_model.pkl`, `output_KNN.hdf5`, `point_NZ.hdf5`, `point_samples.hdf5`, `sampletest.hdf5`, and `varinf_NZ.hdf5`. Do these need to be saved and version controlled in the repo? Or are they just the outputs of the examples? I removed them from the repo because I think they're the latter, but if they're the former, I can restore them!
2. The extended errors require that the galaxy dataframe includes columns called "major" and "minor". I noted this in the docstring. But are these the semi-major and minor axes? Or the full major and minor axes? It looks like you are using them on the same footing as the PSF FWHM in the "auto" method, so I think they must be the full axes? 
3. If these are the full axes, I think the GAaP calculation might be missing a factor of 2? There you convert the PSF FWHM to a Gaussian sigma, and multiply the "major" and "minor" values by the factor that converts a half-IQR to a Gaussian sigma.

Basically, to summarize 2 and 3 above: in the "auto" method, it looks like you are treating the "major" and "minor" values like they are diameters, but in the "gaap" method, you are treating them like they are radii.

Does that make sense? Maybe I am misunderstanding the methods.

But other than these few points, great work! We'll get this merged very soon!

yanzastro:
Hi @jfcrenshaw, thank you for the report! Let me reply to each of them:

1. those output files are generated during a previous test. They are not needed but I forgot to delete them and thank you for pointing it out!
2,3. yes I did mixed up some axes and semi-axes. The cosmoDC2 catalog provides half-light *radius* so everything in the "auto" case is radius or semi-axes; for the "GAAP" case, the maximum aperture size A_max and A_min are given as FWHM, so I convert them into sigma first in line 388 and 389. I converted HL radius to sigma of semi-major and semi-minor radius in line 383 (a fact of 2 was missing here)

jfcrenshaw:
Great! I think we're almost there. Just a few more questions:

1. https://github.com/LSSTDESC/RAIL/blob/a1a3fe199f0cd7a879ec0de20c48d55609d21fa1/rail/creation/degradation/lsst_error_model.py#L383
What is the origin of the 1/1.35 factor here? It looks like the factor that converts a full IQR to a Gaussian sigma? If that's what it is, we should change the comment next to it. But it also makes me worry that we are mixing up radii and diameters again. Since the half-light radius is obviously a radius, shouldn't we still have the half-IQR to Gaussian factor, and not the full-IQR to Gaussian factor? (But I may be totally misinterpreting this factor!)

2. In the "auto" method, we are still using the PSF size and the majors/minors on the same footing. But the PSF size is the FWHM which is a diameter, and the majors/minors are radii. Should we divide the PSF size by 2 to put them on the same footing? Or is it correct as-is?

Again, sorry if my comments above are off base, since I don't know the specifics of these models!

yanzastro:
@jfcrenshaw you are absolutely correct! I confused myself on "half" "semi-" "radius" etc.

For 1, the conversion should be 1/0.6745 (see https://en.wikipedia.org/wiki/Interquartile_range). It converts the half IQR (i.e. the HL radius from the catalog to my understanding) to Gaussian sigma;

For 2, yes I forgot to convert PSF to radii (sigma). Now I update it by multiplying with the factor `fwhm_to_sigma = 1 / 2.355`

jfcrenshaw (approved):
Assuming all the tests pass again, this is ready to merge!

yanzastro:
@jfcrenshaw great! thanks so much!

PR #240, opened 2022-08-31
Add garbage collection to loop in CatEstimator

joezuntz:
I mentioned this in slack - BPZ memory usage ends up as 16GB without a manual call to garbage collection.  I'm not clear why, but this seems to help and shouldn't hurt.


sschmidt23 (approved):
Looks good, thanks Joe, approved and will merge.

PR #247, opened 2022-09-06
Issue/236/obs cond

### LINKED ISSUE #236: Add a degrader to modify magnitude error depending on observing conditions
We want to create a degrader package which sample a set of (correlated) observing conditions and compute the corresponding magnitude error for each galaxy.

Input files: observing conditions (e.g. OpSim outputs or supreme maps) in the form of healpix pixels over the survey footprint (e.g. DC2 wfd, or the full LSST footprint from the up-to-date OpSim version);

Functions: randomly assign galaxies into the pixels, and compute the corresponding magnitude error according to [Ivezic et al. 2019](https://arxiv.org/pdf/0805.2366.pdf) (This is alreay implemented in [lsst_error_model.py](https://github.com/LSSTDESC/RAIL/blob/main/rail/creation/degradation/lsst_error_model.py));

Output: a set of magnitude errors for each galaxy, which is then passed to generate the observed magnitude.

hangqianjun:
This is draft pull request for the ObsCondition degrader.


rail/creation/degradation/observing_condition_degrader.py
```diff
@@ -0,0 +1,383 @@
...
+from ceci.config import StageParameter as Param
+
+#### To do list:
+### - Other formats than fits: .hs (needs healsparse), and .npz (needs rubin)
+### - Should allow input argument for all lsst error model
+
+
+class ObsCondition(Degrader):
+    """
+    Degrader that generates magnitude errors using LSSTErrorModel
```
jfcrenshaw: Make the first line of the docstring a summary of the degrader. So it should mention that this degrader calculates spatially-varying photometric errors using an input map.

rail/creation/degradation/observing_condition_degrader.py
```diff
@@ -0,0 +1,383 @@
...
+    random_seed:
+        A random seed for reproducibility.
+        
+    Example config_file: see example_obs_config
+    """
+    
+    name = 'ObsCondition'
+    config_options = Degrader.config_options.copy()
+    config_options.update(
+        obs_config_file=Param(
```
jfcrenshaw: Rather than pointing to a config file, actually put the parameters into the class.

jfcrenshaw: You could use 
```python
nside
mask
nYrObs
map_dict
```
where `map_dict` is a dictionary that contains
```python
{
    "m5": {"u": path, ...}, 
    "theta": {"u": path, ...},
}
```

OR you could directly make `m5`, `theta`, etc. parameters

rail/creation/degradation/observing_condition_degrader.py
```diff
@@ -0,0 +1,383 @@
...
+        # loop over each pixel
+        pixel_cat_list = []
+        for pixel, pixel_cat in catalog.groupby("pixel"):
+            # get the observing conditions for this pixel
+            obs_conditions = self.get_pixel_conditions(pixel)
+
+            # calculate photometric errors for this pixel
+            errorModel = LsstErrorModel(**obs_conditions)
+            
+            # reset the index
```
jfcrenshaw: Why are you doing this with the indices? Does the error model require that the indices be sequential?

rail/creation/degradation/observing_condition_degrader.py
```diff
@@ -0,0 +1,383 @@
...
+        # assign each galaxy to a pixel
+        catalog = self.assign_pixels(catalog)
+
+        # loop over each pixel
+        pixel_cat_list = []
+        for pixel, pixel_cat in catalog.groupby("pixel"):
+            # get the observing conditions for this pixel
+            obs_conditions = self.get_pixel_conditions(pixel)
+
+            # calculate photometric errors for this pixel
```
jfcrenshaw: This is actually just creating the error model. The errors aren't calculated until line 357 below

jfcrenshaw:
@hangqianjun this is looking great! Just a few comments below, plus:
- you should add some small map files that can be used to test the degrader during the Github action
- I think you added all of the yaml's in `tests/data/test_ObsCond` so that you could run tests with different configurations of the degrader? If so, then if you move the parameters explicitly into the degrader like I suggested above, we can remove all of these yamls

jfcrenshaw (changes requested):

rail/creation/degradation/observing_condition_degrader.py
hangqianjun: I reset the index because the DataFrame for magnitude errors from LSSTErrorModel has an indexing of 0,1,2.., whereas the looped pixel_cat DataFrame doesn't. Directly concatenate them doesn't return the desired object, but a table with either NaN in magnitudes or NaN in magnitude error. Maybe there is an easier way to deal with this?

rail/creation/degradation/observing_condition_degrader.py
hangqianjun: I have changed this to `# creating the error model for this pixel` and then moved the `# calculate the error model for this pixel` to the line above calling `errorModel()`.

rail/creation/degradation/observing_condition_degrader.py
hangqianjun: I have changed the description to:

     ###Photometric errors based on observation conditions

    This degrader calculates spatially-varying photometric errors 
    using input survey condition maps. The error is based on the 
    LSSTErrorModel from the LSST Overview Paper:
    https://arxiv.org/abs/0805.2366
    
    Example:
    ...

rail/creation/degradation/observing_condition_degrader.py
hangqianjun: This has now been implemented.

hangqianjun:
`observing_condition_degrader.py` has now been updated and the tests are merged into `test_degraders.py`. I have added the dependences for `healpy` and `photerr` in `setup.py`.

rail/creation/degradation/observing_condition_degrader.py
jfcrenshaw: I fixed the indexing issues in PhotErr, so @hangqianjun has removed these lines from the code

jfcrenshaw (approved):
Everything looks good to me!

PR #250, opened 2022-09-12
Issue/238/reorganize

### LINKED ISSUE #238: Stray installed directories
When installing `pz-rail-hub` I get several packages which look like they should be sub-packages installed directly to the base `site-packages` directory.  

They are:
```
docs
examples
scripts
tests
```
From looking at them I assume they shouldn't be installed at all.

### LINKED ISSUE #244: Paths to default files
Various stages (at least the ones listed below) have default parameters that assume we are running from the main RAIL directory. With the new installation using pip that's no longer going to be correct.  We should update the paths to their installed location.

I believe the official way to find resources like this that are installed with pip is to use the `importlib.resources` module. I think that should find them whether we have installed or are running directly from the RAIL dir.  The alternative is to use the `__file__` variable if I've misunderstood this.
This will have to wait until after #238 since that will presumably involve moving the files.

Affected stages:
```
GridSelection
SpecSelection
Inform_BPZ_lite
BPZ_lite
Inform_DelightPZ
delightPZ
```

### LINKED ISSUE #245: Simplify pip install options
When things were more complicated with Delight, FZBoost, etc... we created install options for .[base], .[all], etc..., now that we've moved those packages out we should simplify this a bit and change the default install to grab all necessary packages (a few like pzflow used by PZFlowPDF are not installed by default with a simple `pip install pz-rail`).

We can discuss a bit whether we want any remaining options, or whether we should basically just have everything in pip install .[all] basically be the default option.

sschmidt23:
Leaving as a draft for now, I think this is mostly done, but I need to make sure I don't have duplicate copies of files and such.  This PR addresses several interrelated issues, namely 
#238 (package_data files are now moved to `src/rail/examples/` rather than top-level `examples/`)
#245 (I eliminated the various install options, base, all, etc..., so we how just have the one standard install)
#244 (already partially fixed by addition of RAILDIR, but going through and checking default file locations for GridSelection and SpecSelection to use RAILDIR and new file locations for defaults).  BPZ and Delight changes will need to be made in rail_bpz and rail_delight separate from this PR.

For #245, in eliminating the other install options, I added sklearn and pzflow to the overall package requirements.  As pzflow is also used by one of the estimators, and sklearn is pretty standard, this seems uncontroversial to me, but mentioning it here just in case.

Will assign reviewers once I finish file location checks and run this on a fresh install on my other computer.


sschmidt23:
Ok, I think this is ready now, I will have to make separate PRs for rail_bpz and rail_delight after this is approved, as the move will break a couple of default config file location conventions in those repos I believe.

A thing to check for reviewers: after you pip install RAIL, check the site-packages directory where the files are copied and make sure that examples, tests, etc... directories are not copied there outside of the RAIL package.  This works correctly for me, but outside confirmation is always good.

eacharles (approved):
Installed it with pip install . and pip install -e ., ran unit tests, both worked.   Made sure the installed files end up in the right place.    Ran notebooks using the documentation scripts. everything looked good.

PR #251, opened 2022-09-13
fix version

eacharles:


sschmidt23 (approved):
This is now properly picking up the version and seems to fix the creation of the sdist, approved

PR #254, opened 2022-09-19
fix photometric_realization_demo

yanzastro:
thanks for removing the redundant lines! I think the issue is fixed now.


eacharles (approved):
Thanks

Issue #257, opened 2022-09-20
Request to rename sklearn dependency to scikit-learn


heather999:
Use scikit-learn rather than sklearn in the toml file. The official package name is scikit-learn and in conda-forge, that's the name that is available, not sklearn. See: https://pypi.org/project/sklearn/  that is a dummy pypi project
Why does it matter? Because if I wanted to set up an environment that uses the conda-forge scikit-learn and then install RAIL using pip - it goes off and installs the dummpy sklearn pypi package which is confusing.

sschmidt23:
closed by #258

PR #258, opened 2022-09-20
use scikit-learn rather than sklearn

heather999:
avoids pulling in dummy pypi project - see #257


sschmidt23 (approved):
Thanks, Heather, I'll need to fix this same issue in pz-rail-flexzboost as well, as we have the same issue there.

PR #262, opened 2022-09-26
Issue/125/add brier metric

### LINKED ISSUE #125: Add Brier Score metric to evaluate
Implement Brier Score metric in RAIL.evaluate.

drewoldag:
Added the basic Brier metric assuming an input of Numpy arrays (or Python lists). Also introduced a suite of unit tests for the new metric to get full code coverage. Checked code style with pylint.

### DIFF
--- a/src/rail/evaluation/metrics/brier.py
+++ b/src/rail/evaluation/metrics/brier.py
@@ -41,7 +41,7 @@ class Brier(MetricEvaluator):
     def _manipulate_data(self):
         """
         Placeholder for data manipulation as required. i.e. converting from
-        qp.ensamble objects into np.array objects.
+        qp.ensemble objects into np.array objects.
         """
 
         # Attempt to convert the input variables into np.arrays
@@ -79,7 +79,7 @@ class Brier(MetricEvaluator):
         """
 
         # If inputs are 1-d arrays adjust axis and retained output dimensions.
-        # Note: It's safe to check just self._prediction, becuase in
+        # Note: It's safe to check just self._prediction, because in
         # _validate_data, we raise an exception if _prediction and _truth do not
         # have the same shape.
         AVERAGE_ALONG_ROWS = 1
--- a/src/rail/evaluation/metrics/brier.py
+++ b/src/rail/evaluation/metrics/brier.py
@@ -1,0 +1,28 @@
+import numpy as np
+from .base import MetricEvaluator
+
+class Brier(MetricEvaluator):
+    """ Brier score """
+    def __init__(self, prediction, truth):
+        """ Class constructor """
+        super().__init__(None)
+
+        self._prediction = np.array(prediction)
+        self._truth = np.array(truth)
+
+    def evaluate(self) -> float:
+        """
+        Evaluate the Brier score
+
+        Parameters
+        prediction: numpy.ndarray, float
+            predicted class probabilities
+        truth: numpy.ndarray, int
+            true classes mask
+
+        
+
+        Based on https://en.wikipedia.org/wiki/Brier_score#Original_definition_by_Brier
+        """
+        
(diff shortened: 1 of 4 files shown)


src/rail/evaluation/metrics/brier.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+
+        self._manipulate_data()
+        self._validate_data()
+        return self._calculate_metric()
+
+    def _manipulate_data(self):
+        """
+        Placeholder for data manipulation as required. i.e. converting from
+        qp.ensamble objects into np.array objects.
```
delucchi-cmu: ensemble?

delucchi-cmu (commented):
Just a minor nit. Otherwise LGTM.

src/rail/evaluation/metrics/brier.py
drewoldag: Yep, good call. Also prompted me to install a spell checker (Code Spell Checker) in VS code, which immediately found another spelling mistake 🤦

src/rail/evaluation/metrics/brier.py
```diff
@@ -0,0 +1,94 @@
...
+        # Note: It's safe to check just self._prediction, because in
+        # _validate_data, we raise an exception if _prediction and _truth do not
+        # have the same shape.
+        AVERAGE_ALONG_ROWS = 1
+        KEEP_DIMS = False
+        if self._prediction.ndim == 1:
+            AVERAGE_ALONG_ROWS = 0
+            KEEP_DIMS = True
+
+        return np.mean(
```
jbkalmbach: This should be the mean of the sum of the squared value along each row.

src/rail/evaluation/metrics/brier.py
```diff
@@ -0,0 +1,94 @@
+import numpy as np
+from .base import MetricEvaluator
+
+class Brier(MetricEvaluator):
+    """ Brier score """
+    def __init__(self, prediction, truth):
+        """
+        Parameters
+        ----------
+        prediction: NxM array, float
```
jbkalmbach, line 11: Since each row should be a probability distribution should we have a quick check that they each sum to 1? If it is understood then maybe it's not necessary but it might be helpful to add a check and something here mentioning that.

src/rail/evaluation/metrics/brier.py
jbkalmbach: The output for any size NxM input should be a single score with a minimum value of 0 and a maximum value of 2.

jbkalmbach (changes requested):

src/rail/evaluation/metrics/brier.py
drewoldag: We just had a quick sync in the meeting room, and thanks for the additional clarification about the way the output will be used. It seems to be the "original definition": https://en.wikipedia.org/wiki/Brier_score#Original_definition_by_Brier

I'll make the required modifications, and update the tests.

src/rail/evaluation/metrics/brier.py
drewoldag: This seems like a good idea to me. I'll likely include that in the _validate_data method.

drewoldag: Updated the docstring here to specify that each row should sum to 1. Also implemented a check in _validate_data that will log a warning if there are rows that don't sum to 1.

src/rail/evaluation/metrics/brier.py
drewoldag: I updated the metric calculation based on your guidance and have pushed a commit for this. If you have a chance to take another look, it would be very much appreciated :)

jbkalmbach (approved):
Updated metric looks good to me. Thanks for the help adding this in!

PR #286, opened 2022-11-17
First working version with NZDir running in parallel

### LINKED ISSUE #209: Add parallelization to the summarizers

joselotl:
I wrote a first version of NZDir working in parallel. It uses an iterator over the photometric data and later combines the resulting histograms to a single one.

I'm wondering if the same structure used here can be used in other similar summarizers.
@sschmidt23 Which summarizers are similar in structure to NZDir?
@joezuntz Have you started working on somocluSOM? Otherwise I think I can use the same algorithm there?

### DIFF
--- a/examples/estimation/nzdir_as_pipeline.ipynb
+++ b/examples/estimation/nzdir_as_pipeline.ipynb
@@ -1,0 +1,304 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "id": "e324bb13",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "import rail\n",
+    "import qp\n",
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "import matplotlib.pyplot as plt\n",
+    "%matplotlib inline\n",
+    "from rail.estimation.algos.NZDir import Inform_NZDir, NZDir\n",
+    "from rail.core.data import TableHandle, QPHandle\n",
+    "from rail.core.stage import RailStage"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "8d2352e5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "DS = RailStage.data_store\n",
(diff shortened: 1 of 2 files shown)


sschmidt23:
Running ceci via the `nzdir_as_pipeline.ipynb` notebook on cori is still acting up, but running via a submitted batch job with ceci works as expected, so I think everything looks good.

But, before I approve, @joselotl you made a comment this morning about something not working if we were not using a file or something like that?  What was the problem? And should we fix it in this PR, or should we approve and merge this and then create a separate issue?

sschmidt23 (approved):
I didn't approve this last month due to comment about a separate issue, I'll approve now, if there is a separate problem we can create a separate issue to address it.

PR #287, opened 2022-11-17
Issue/285/som hc

### LINKED ISSUE #285: Adding hierarchical clustering and quality control into SOM summarizer
Currently, the SOM summarizer is analyzed based on cells (for example, N(z) are synthesized across cells). However, this might be affected by sample noise. In addition, some cells may not contain spectroscopic sample, so the sources in these cells are not represented and need to be removed.

Following https://www.aanda.org/articles/aa/abs/2020/05/aa36782-19/aa36782-19.html, it would be better to group cells into hierarchical clusters and summarize N(z) based on these clusters. By doing so, we might sacrifice some accuracy in N(z) estimation but will have more photometric sources represented. Therefore we need to find the number of clusters that will balance this trade-off.

yanzastro:
I added hierarchical clustering function into the somoclu summarizer and add some example cells in the corresponding example notebook.


src/rail/estimation/algos/somocluSOM.py
```diff
@@ -386,48 +393,74 @@ def run(self):
...
+        phot_cluster_set = set(phot_som_clusterind)
+        spec_cluster_set = set(spec_som_clusterind)
+        uncovered_clusters = phot_cluster_set - spec_cluster_set
+        bad_cluster = dict(uncovered_clusters=np.array(list(uncovered_clusters)))
+        print("the following clusters contain photometric data but not spectroscopic data:")
+        print(uncovered_clusters)
+        useful_clusters = phot_cluster_set - uncovered_clusters
+        print(f"{len(useful_clusters)} out of {n_clusters} have usable data")
+
+        n_eff = np.sum(pweight) ** 2 / np.sum(pweight**2)
```
sschmidt23: Can you put a short comment, either inline here or in the main big docstring comment, giving the definition of n_eff, n_eff_p_nu, and the ratios, as well as how you expect to use them?  I see them used in the demo notebook, but there isn't much description of what they are or why they are important.

sschmidt23 (approved):
I left one comment about adding a bit more explanation of n_eff and n_eff_p_num, if you can add that it might be helpful to users to have more documentation in the code.  But, otherwise, looks good.

One more comment: on running the demo notebook it looks like the training stage for somoclu spawns a bunch of processes on my home machine, memory usage briefly jumps up to like 25 GB, then drops.  However, it looks like the processes persist and still have about 1.5GB of memory each even after the job completes.  So, there may be a memory leak issue with somoclu.  If you see this behavior maybe think about creating an issue for them.

sschmidt23:
Oh, and one more thing, even though I approved this: do we actually need to include `output_SOMoclu_model.pkl`?  RAIL is getting rather bloated, if the file is not needed then we should delete it from the repo before you merge this PR.

yanzastro:
Hi @sschmidt23 and sorry for the late reply. I forgot to check my RAIL email folder again. Here I'll reply to all your comments:

1. I'll add the docstring defining n_eff, etc tomorrow;
2. I'll also check the memory issue and raise an issue for somoclu;
3. I don't think we need output_SOMoclu_model.pkl as it is the output of a trained SOM. I'll delete and push the repo

sschmidt23:
I just noticed that a couple changes that I made to get the notebook running and some linting fixes did not actually get pushed to the repo for some reason, I'll test re-running the notebook and apply the needed changes now.

sschmidt23 (approved):
Looks good, thanks Ziang

PR #290, opened 2022-11-28
Issue/268/dspsinterface

### LINKED ISSUE #268: Adding interface in rail.creation to DSPS code
The aim is to add an interface in RAIL to the stellar population synthesis code "DSPS".

torluca:
This is a pull request to merge the implementation of an interface to DSPS into the main branch.
The files that have been added are in the src/rail/creation/galaxy_modelling folder.
I added a demo notebook on how to run DSPS (/examples/creation/dsps_sed_demo.ipynb).
Unit tests are still missing and taking now most of the time.


torluca:
I realised there is a bug in one of the unit test I implemented, I'm going to fix it by tomorrow

torluca:
Fixed. @yanzastro if everything looks ok to you than we can proceed to merge it into the main branch

yanzastro (commented):
I have reviewed this PR. The code looks good to me. But the notebook generates a lot of .npy files, I think you should remove them from the repo (Like SFH.npy. Generally, the repo shouldn't include files generated by example codes).

Also, I didn't look into DSPS package very closely since I'm not very familiar with SED models. It might be helpful to invite another reviewer for this part.

torluca:
> I have reviewed this PR. The code looks good to me. But the notebook generates a lot of .npy files, I think you should remove them from the repo (Like SFH.npy. Generally, the repo shouldn't include files generated by example codes).
> 
> Also, I didn't look into DSPS package very closely since I'm not very familiar with SED models. It might be helpful to invite another reviewer for this part.

Do you suggest generating the *.npy files content on the fly?

Because those files are generated externally and only read by unit tests

yanzastro:
No, we should keep the necessary files generated externally (like `ssp_spec_flux_lines.npy` and `ssp_spec_wave.npy` ), but I noticed that some files are generated from the code (the example notebook) itself (like `metallicity_grid.npy`, `age_grid.npy`, etc). I think these files should be removed from the repo to save space. It will be generated by the user locally when they run the code anyways.

torluca:
> No, we should keep the necessary files generated externally (like `ssp_spec_flux_lines.npy` and `ssp_spec_wave.npy` ), but I noticed that some files are generated from the code (the example notebook) itself (like `metallicity_grid.npy`, `age_grid.npy`, etc). I think these files should be removed from the repo to save space. It will be generated by the user locally when they run the code anyways.

I see, let me remove some of them and see if the tests still work

torluca:
@yanzastro I removed from the testdata all the unnecessary files that are not needed as defaults by the unit test. I also commented the lines in the notebook where the files are saved inside rail, so users do not pile up rail/testdata

yanzastro:
Hi Luca, I guess I didn't make my point clear. I didn't mean that you should change your code not to generate those files. I meant that those files shouldn't be pushed into the repo. The code looks good to me now, the only thing you need to do is to `gitignore` the generated files when you push the PR. By doing this, the RAIL repo will not include these files, but the users can generate them locally by running the notebook.

In any case, I see that you removed the code generating those files, and it passed the code test. If you are comfortable with the current code (without saving those files), I can proceed to merge it.

torluca:
By gitignoring those files, then the unit tests need to generate them on the fly to pass, am I right?

yanzastro:
Yes I think so.

torluca:
I see, unfortunately there are certain files that cannot be generated on the fly, because they are generated from an external code (namely FSPS). So at least some of them need to be saved into the testdata directory and git pushed otherwise the unit test will fail. I will try to minimise their number further.

yanzastro:
Thanks! For those files generated from external codes, you can save and push them to the PR.

torluca:
Hi @yanzastro , I have rewritten the unit tests and the demo notebook. Now the files are generated on the fly, apart from the three externally generated ones that are in rail/examples/testdata. The demo notebook now contains the code to generate the file needed to run DSPS. Thanks for your help, the code is cleaner now.

yanzastro (approved):

PR #293, opened 2022-12-12
change flow= to model= in FlowCreator stages

### LINKED ISSUE #292: Pipe_Example notebook fails to run out of the box
When running **RAIL/examples/core/Pipe_Example.ipynb**, even from a completely fresh installation of RAIL, the cell that calls `pipe.initialize(dict(flow=flow_file), dict(output_dir='.', log_dir='.', resume=False), None)` throws:
```
ValueError: 
Some required inputs to the pipeline could not be found,
(or possibly your pipeline is cyclic):

Stage flow_engine_test is missing input(s): model
Stage lsst_error_model_test is missing input(s): output_flow_engine_test
Stage col_remapper_test is missing input(s): output_lsst_error_model_test
Stage table_conv_test is missing input(s): output_col_remapper_test
```

#### Steps to reproduce:
```zsh
conda create -n freshrail pip python=3.9
conda activate freshrail
conda install healpy
git clone https://github.com/LSSTDESC/RAIL.git
cd RAIL
pip install -e '.[all]'
conda install -n freshrail ipykernel --update-deps --force-reinstall
```
Note: I was still getting the hang of Python notebooks with Github in my last PR, and had accidentally left an extra comma in the JSON when removing metadata. So, line 314 is currently `},` when it should be `}`. I'll tidy this in my next PR. 

You can either manually remove the comma on line 314, or grab the [version of the notebook that predates my edits](https://github.com/LSSTDESC/RAIL/blob/66bb416785b9f9785c43a9674b91fbb36f2d2553/examples/core/Pipe_Example.ipynb) and then find/replace the deprecated `FlowEngine` with `FlowCreator` (but make sure it’s a case sensitive search - we want to keep `flowEngine` as it is).

I was able to recreate the bug identically in both versions of the notebook.

sschmidt23:
fixes the name change in specifying FlowCreator from `flow=` to `model=`, notebook runs fine for me after this small change

### DIFF
--- a/examples/core/Pipe_Example.ipynb
+++ b/examples/core/Pipe_Example.ipynb
@@ -6,8 +6,8 @@
    "source": [
     "# RAIL Pipeline example notebook\n",
     "\n",
-    "author: Eric Charles\n",
-    "last run successfully: March 16, 2022\n",
+    "author: Eric Charles<br>\n",
+    "last run successfully: Dec 12, 2022<br>\n",
     "\n",
     "This notbook shows how to:\n",
     "1. Build a simple rail pipeline interactive, \n",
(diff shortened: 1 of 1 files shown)


olivialynn (approved):
Fantastic, thanks!

PR #295, opened 2023-01-04
switch np.int to just int

sschmidt23:
A trivial PR to change the now deprecated `np.int` to `int` in somocluSOM.py which was causing a test failure starting with numpy v1.24.


drewoldag:
This looks good to me.

drewoldag (approved):

PR #301, opened 2023-02-28
Allowing notebook workflow to open PRs

olivialynn:
Notebook updater now opens a PR (original commit action doesn't work with a protected branch); also added back the cron job. Running it on the first of each month, so automated PRs don't introduce too much clutter.


eacharles (approved):
Looks good, thanks.

PR #306, opened 2023-03-08
Issue/304/cleanup

sschmidt23:
For a minor clean up, this PR just deletes two old pickled model files and a duplicate copy of the HDFN prior.  One of the tests did use the demo_snn.pkl file that was deleted, but it's only a check on opening a pickled model file, so I changed that to instead read in the remaining copy of CWW_HDFN_prior.pkl.


sschmidt23:
And we should create a new tagged version of RAIL once this is merged to resolve the `sklearn` deprecation that is now being enforced by PyPI

eacharles (approved):
Thanks Sam.

PR #308, opened 2023-03-10
update files

### LINKED ISSUE #307: Speed up get_bmus function in somoclu summarizer
The current `get_bmus` function of somocluSOM summarizer is not efficient and can be speeded up.

yanzastro:


src/rail/estimation/algos/somocluSOM.py
```diff
@@ -44,30 +44,45 @@ def _computemagcolordata(data, ref_column_name, column_names, colusage):
...
+    split: an integer specifying the size of data chunks when calculating the distances.
+    returns: The the dot product of the codebook and the data.
+    rtype: 2D numpy.array
+    """
+
+    d = data
+
+    codebookReshaped = som.codebook.reshape(
+        som.codebook.shape[0] * som.codebook.shape[1], som.codebook.shape[2])
+    parts = np.array_split(d, 200, axis=0)
```
sschmidt23: where does this value of 200 come from?  And why do you split into 200 parts and loop over them when computing distances?

yanzastro: Yes. I split the array into 200 slices and loop over. I tried several numbers and found 200 seems optimal (in the sense of speed) I will update it so users can choose the number to optimise the speed.

sschmidt23:
I ran the demo notebook with the old and updated versions of the functions that you added. Strangely, the CPU time used looks to be about the same old vs updated, but the wall time is reduced for your new function methods by around a factor of four.  I'm not sure why the old way had what looks like random hangs built in, but it does look like it's working and the code now runs faster.  So, I'll approve once you add a note about the slicing in the one comment in the code, as things look good.

yanzastro:
Hi @sschmidt23, thanks for reviewing the PR! Basically what I did is to add our own function to calculate the distance between data vector and SOM weight vector because the built-in function in `somoclu` is slow. Here is the original function in somoclu (see line 556 in https://github.com/peterwittek/somoclu/blob/7c784488c74954d0e24915a00c83a5b2864ff1b8/src/Python/somoclu/train.py):
`    def get_surface_state(self, data=None):

        if data is None:
            d = self._data
        else:
            d = data

        codebookReshaped = self.codebook.reshape(self.codebook.shape[0] * self.codebook.shape[1], self.codebook.shape[2])
        parts = np.array_split(d, 200, axis=0)
        am = np.empty((0, (self._n_columns * self._n_rows)), dtype="float64")
        for part in parts:
            am = np.concatenate((am, (cdist((part), codebookReshaped, 'euclidean'))), axis=0)

        if data is None:
            self.activation_map = am
        return am`

The distances are stored in the array `am` which is initialized as a zero-size array. The distances are calculated chunk by chunk with the for loop. In each iteration, the code simply concatenates the distances to the old `am` array. This makes the function run pretty slowly.

What I updated is just to define `am` with the size of the data and assign the distances from each loop to the associated part of `am`.

Now I also combine the `get_surface_state` function with `get_bmus` function to reduce redundancy

yanzastro:
Hi @sschmidt23, thanks for reviewing the PR! Basically what I did is to add our own function to calculate the distance between data vector and SOM weight vector because the built-in function in `somoclu` is slow. Here is the original function in somoclu (see line 556 in https://github.com/peterwittek/somoclu/blob/7c784488c74954d0e24915a00c83a5b2864ff1b8/src/Python/somoclu/train.py):
 
```
def get_surface_state(self, data=None):

        if data is None:
            d = self._data
        else:
            d = data

        codebookReshaped = self.codebook.reshape(self.codebook.shape[0] * self.codebook.shape[1], self.codebook.shape[2])
        parts = np.array_split(d, 200, axis=0)
        am = np.empty((0, (self._n_columns * self._n_rows)), dtype="float64")
        for part in parts:
            am = np.concatenate((am, (cdist((part), codebookReshaped, 'euclidean'))), axis=0)

        if data is None:
            self.activation_map = am
        return am

```

The distances are stored in the array `am` which is initialized as a zero-size array. The distances are calculated chunk by chunk with the for loop. In each iteration, the code simply concatenates the distances to the old `am` array. This makes the function run pretty slowly.

What I updated is just to define `am` with the size of the data and assign the distances from each loop to the associated part of `am`.

Now I also combine the `get_surface_state` function with `get_bmus` function to reduce redundancy

sschmidt23 (approved):
I fixed an indent in a quick commit, but this now looks good, thanks Ziang.  I just wanted to make sure one of us didn't get confused as to where that 200 number came from at some point in the future.

Issue #320, opened 2023-03-29
General cleanup questions


olivialynn:
### Inconsistent use of data and configs directories

There are inconsistencies in the existence/use of `data/` and `configs/` directories in each directory in `RAIL/src/rail`.

  | configs in ./ | configs in ./configs/ | data in ./ | data in ./data/
-- | -- | -- | -- | --
RAIL/src/rail/examples/creation | x | yes | x | yes
RAIL/src/rail/examples/estimation | x | x | yes | yes
RAIL/src/rail/examples/goldenspike | yes | x | x | yes

We aso have some stray configs outside of src/rail, such as RAIL/examples/core/[pipe_example_config.yml](https://github.com/LSSTDESC/RAIL/blob/main/examples/core/pipe_example_config.yml).

* Do we have a preference for always using directories, or always avoiding their use? 
* Do we want to strictly keep all config files within src/rail? 

### Pipelines directory

Currently, is has its own .gitignore containing just  `*.*~`, and nothing else in RAIL seems to use it. 

* Is this meant to just be a landing spot for users who also install [rail_pipelines](https://github.com/LSSTDESC/rail_pipelines/)? 
* Do we want to encourage users who install rail and build their own pipelines to place them in that directory?

### Attic

There's an attic directory at [RAIL/examples/goldenspike/attic/](https://github.com/LSSTDESC/RAIL/tree/main/examples/goldenspike/attic).

It seems to create at least a nonzero amount of maintainability work; I've noticed Drew had to [update a couple of the utils](https://github.com/LSSTDESC/RAIL/commit/231dd1312c3afaaa7c3bc01c6225c145e4141342#diff-b5fd1dde7a78617518e00bec7def500393ab2c5bcce9313d3c6b38b853d81139) in his PR deprecating Brier and PIT.

The name attic implies a place for old/unused things, but this seems like a directory that lets the user interact with goldenspike outside of a notebook. 

* Are we keeping it around for users who want more than just a notebook? If so, do we want to rename the directory to something more descriptive?
* Otherwise, do we want to just delete it?

### Misc
* I'm currently working on #182, so skipping over outdated READMEs/html renders
* [RAIL/examples/evaluation/utils.py](https://github.com/LSSTDESC/RAIL/blob/main/examples/evaluation/utils.py) -> do we want a utils file in our examples directory? If so, can we rename this to reflect that?

PR #324, opened 2023-04-05
General cleanup

olivialynn:
This PR addresses topics listed in #320 (data/configs consistency, legacy pipelines dir, attic)

~~Keeping this as a draft while I investigate the dsps problems we've been seeing in the github workflows lately.~~ 
(Edit: CI fixed, this should be good to go.)


src/rail/examples/goldenspike/goldenspike.yml
drewoldag, line 1: Is this file considered a configuration file? And if so does it belong in the `.../goldenspike/configs/` directory?

drewoldag (approved):
All of this looks good. I see that you've cleared out the `.../attic` and removed the `.gitignore` from `.../pipelines`. That's all good. 

I left one question on the `goldenspike.yml` file about whether it is suppose to be in the configs directory or not. 

All in all, this looks good to me.

src/rail/examples/goldenspike/goldenspike.yml
olivialynn: I was on the fence about this too, but it looks to me to be a pipeline file, rather than a config file (and it references the goldenspike config file, as well--a config that references another config seems unlikely to me). Very open to moving it if this is the wrong interpretation.

drewoldag: Got it, yeah, I don't have enough context in how this is used to offer meaningful advice about where it belongs. So I would be in favor of leaving it as.

PR #326, opened 2023-04-07
Add @v0.2.0 to dsps

olivialynn:
Addresses the failure for dsps-related tests that we've been seeing in our CI the past week.

We import dsps from CPAC's GitHub, so I went ahead and added version specifiers for the other two repos we import from them, as well.


drewoldag (approved):
This looks good to me. Took a look at the check steps where dependencies are installed and those look find as well.

PR #329, opened 2023-04-11
Move hardcoded variable names in two estimation stages to config params

### LINKED ISSUE #47: Input data naming in estimation codes

sschmidt23:
Addressing the last vestiges of issue #47 where a seldom used "dummy" photo-z code that just generates random Gaussians and our pathological case that just sets every PDF to the histogram of training values still had some hardcoded names from way back.  A couple minor linting updates like spaces around operators as well.  I tested everything and it all works in a notebook.


drewoldag (approved):
This looks good, thanks!

PR #331, opened 2023-04-12
Issue 330: fix test variable mistake and quant name

sschmidt23:
@eacharles pointed out something obvious that I missed: that the test was actually feeding in ez values rather than specz values, which somehow did not cause a crash until now.  That and changing the name from quant_piecewise to quant in one of the PIT functions now has all of the tests running for me on my local branch.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "minisom",
     "scipy>=1.9.0",
     "pz-hyperbolic-temp",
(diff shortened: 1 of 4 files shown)


sschmidt23:
A couple of tests failed because of the changes going on in the minimal tables_io imports PR. A new tagged version of qp with tables_io[full] should fix this shortly, at which point the tests should pass once we re-launch them.

drewoldag (approved):
Looks good to me

PR #335, opened 2023-04-19
Add pip install wheel to main.yml

### LINKED ISSUE #333: Main.yml fails to build wheel when somoclu is a dependency
In a fresh fork of RAIL, I have:

* a [workflow run](https://github.com/OliviaLynn/RAIL/actions/runs/4727096059/jobs/8387331893) that fails the Install dependencies step (with the pyproject.toml exactly as in RAIL)
* a [workflow run](https://github.com/OliviaLynn/RAIL/actions/runs/4727116515/jobs/8387376407) that passes the Install dependencies step (and fails the somoclu tests, because I've taken somoclu out of the pyproject.toml)

Going to keep looking into this, but if anyone has any insight, that would be fantastic.

_(For the record: I did change main.yml to use only python 3.10 to simplify isolating the issue. Also excluded the redundant `pip install numpy` line from main.yml, but I've checked since and the issue remains with or without it.)_

olivialynn:
Fixes #333; lets RAIL run on github actions again. 

Based on the https://github.com/pypa/pip/issues/8559  deprecation thread.

Talked to someone at somoclu and this seemed the best course of action.

(Note, this adds to the explicit `pip install numpy` line we've already needed before installing somoclu: https://github.com/peterwittek/somoclu/issues/165).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,8 +31,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install wheel
-        pip install numpy
+        pip install wheel numpy # For somoclu
         pip install .[all]
         pip install flake8 pytest pytest-cov mockmpi pytest-timeout
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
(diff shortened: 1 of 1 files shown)


drewoldag (approved):
Looks good.

PR #337, opened 2023-04-24
Mechanical updates

delucchi-cmu:
This is the result of running formatting tools, with no manual modifications:

```
python3 -m black tests/
python3 -m isort tests/
```


eacharles (approved):
This looks good, thanks for doing it.  
Could you warn folks on the pz-rail channel that there is a re-formatting PR in the works that that they will probably want to any development branches that they are working on before they start making lots of changes to the unit tests.  Thanks

PR #340, opened 2023-04-25
Mark slow tests and run separately.

delucchi-cmu:
This registers a new mark, `@pytest.mark.slow`, that can be used to annotate test cases that are known to be slow and can be skipped for quick unit testing.

The workflow is modified to first run any tests NOT marked with slow, and on success, run the remaining slow tests. The coverage from the two runs is combined, so this does not result in any change to overall test coverage.

Additional tests will be marked as slow in upcoming PRs, once later "fast" tests no longer rely on state produced by earlier "slow" tests.


eacharles (approved):
Nice, thanks!

PR #351, opened 2023-04-26
Define a new MPI communicator when data set is small

### LINKED ISSUE #275: MPI run with small datasets causes a crash
At least on cori, running  an estimator via a batch job where you have a total number of galaxies that is smaller than the chunk size times the number of processors that you are running on causes a crash, as it sets up N processes, but at least one of those does not receive any data from the iterator, but it still tries to run `finalize_write` and does not end up producing an output file.  A portion of the error message:

```
    PipelineStage.main()
  File "/global/homes/s/schmidt9/.local/cori/3.9-anaconda-2021.11/lib/python3.9/site-packages/ceci/stage.py", line 469, in main
    stage.execute(args)
  File "/global/homes/s/schmidt9/.local/cori/3.9-anaconda-2021.11/lib/python3.9/site-packages/ceci/stage.py", line 604, in execute
    self._output_handle.finalize_write()
AttributeError: 'NoneType' object has no attribute 'finalize_write'
```

joselotl:
When data set is small, define a new communicator that only has the needed processes to compute the current model. This is written in input_iterator so that it works with any algorithm using iterators.

### DIFF
--- a/src/rail/core/stage.py
+++ b/src/rail/core/stage.py
@@ -6,6 +6,7 @@ from ceci import PipelineStage, MiniPipeline
 from ceci.config import StageParameter as Param
 from rail.core.data import DATA_STORE, DataHandle
 
+from math import ceil
 
 class StageIO:
     """A small utility class for Stage Input/ Output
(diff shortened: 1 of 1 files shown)


joselotl:
@eacharles As we discused in the tag-ups, I didn't include the unit tests for this. It would need to use mpi in the github actions.

joselotl:
@eacharles @sschmidt23 This PR is ready to merge, it only needs an approving review

eacharles (approved):
Thanks!!

PR #354, opened 2023-04-28
Addressing the last few remaining places where randoms aren't using default_rng and/or a configurable seed

### LINKED ISSUE #41: managing random seeds uniformly
the np.random.seed call within the current estimation algorithms is probably not the way to go, as it does not allow the user to set it. Suggestion : add seed=87 in the `__init__` of the concerned algorithms, and add a `set_seed` in the base class

sschmidt23:
addressing #41, I looked for instances where we were not using numpy.random.default_rng, and/or were not setting the seed from a config parameter for the stage.  With all stages that use a random number generator, using default_rng and a configurable seed should allow us to isolate the effects of the rng to the particular stage, and easily change via the config parameter to test effects of the randoms on stage performance.

### DIFF
--- a/src/rail/estimation/algos/knnpz.py
+++ b/src/rail/estimation/algos/knnpz.py
@@ -106,8 +106,8 @@ class Inform_KNearNeighPDF(CatInformer):
         trainszs = np.array(training_data[self.config.redshift_column_name])
         colordata = _computecolordata(knndf, self.config.ref_column_name, self.config.column_names)
         nobs = colordata.shape[0]
-        rng = np.random.default_rng
-        perm = rng().permutation(nobs)
+        rng = np.random.default_rng(seed=self.config.seed)
+        perm = rng.permutation(nobs)
         ntrain = round(nobs * self.config.trainfrac)
         xtrain_data = colordata[perm[:ntrain]]
         train_data = copy.deepcopy(xtrain_data)
--- a/src/rail/estimation/algos/randomPZ.py
+++ b/src/rail/estimation/algos/randomPZ.py
@@ -36,7 +36,7 @@ class RandomPZ(CatEstimator):
         pdf = []
         # allow for either format for now
         numzs = len(data[self.config.column_name])
-        rng = np.random.default_rng(seed=self.config.seed)
+        rng = np.random.default_rng(seed=self.config.seed + start)
         zmode = np.round(rng.uniform(0.0, self.config.rand_zmax, numzs), 3)
         widths = self.config.rand_width * (1.0 + zmode)
         self.zgrid = np.linspace(self.config.rand_zmin, self.config.rand_zmax, self.config.nzbins)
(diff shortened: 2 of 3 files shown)


aimalz (approved):
These all look good, thanks for catching and fixing them! _EDIT: Actually, does it matter if `random_seed` isn't explicitly in the init unpacking of the config parameters in two of them?_

src/rail/estimation/algos/randomPZ.py
```diff
@@ -35,7 +36,8 @@ def _process_chunk(self, start, end, data, first):
         pdf = []
         # allow for either format for now
         numzs = len(data[self.config.column_name])
-        zmode = np.round(np.random.uniform(0.0, self.config.rand_zmax, numzs), 3)
+        rng = np.random.default_rng(seed=self.config.seed)
```
drewoldag: This is probably fine, but it stood out a little. The goal here is that every time `_process_chunk` is called, it would use the same random value, not a new random value for each call to `_process_chunk`, correct?

src/rail/estimation/algos/knnpz.py
```diff
@@ -106,8 +106,8 @@ def run(self):
         trainszs = np.array(training_data[self.config.redshift_column_name])
         colordata = _computecolordata(knndf, self.config.ref_column_name, self.config.column_names)
         nobs = colordata.shape[0]
-        rng = np.random.default_rng
-        perm = rng().permutation(nobs)
+        rng = np.random.default_rng(seed=self.config.seed)
```
drewoldag, line 109: No need to update, but this file is using `0` as the default seed - just seems like we could do better 🤷

drewoldag (approved):
This looks ok to me. Thank for taking care of it.

eacharles:
Maybe it makes sets to use the original seed + the chunk number, so each chunk gets a different seed, but in a deterministic way.

Or am I missing something?

-e

sschmidt23:
I hadn't even thought of the chunk number, that's a good point, Eric.  I think you're right, in our current setup, if we have the numpy.random.default_rng(seed=self.config.seed) in the process_chunk function, then it will reset to the *same* seed at the start of each chunk, setting the seed to seed + chunknum fixes that.  I guess the other option would be to set the rng in the init as self.rng so that it's not reset at each call of process_chunk but rather only during the init.

sschmidt23:
Checking through things in RAIL, the only parallelized estimator that uses an rng in `_process_chunk` is randomPZ, I changed the seed initialization to `seed = self.config.seed + start`.  All other uses of a random number generator are in non-parallelized functions, and thus do not need this addition.  However, any future parallelizations that have a random number used in a chunked function will have to similarly initialize (e.g. the open PR on somocluSOM, I'll push a change to that branch now as well).

I'll look through GPz_v1, FlexZBoost, Delight, and BPZ_Lite now to see if I need to make any changes on those repos.

PR #362, opened 2023-05-03
Issue/357/notebook ci

olivialynn:
For LSSTDESC/rail_hub#21 - adding a new workflow for checking all notebooks, so we no longer have to run the workflow that creates a PR and only checks our RTD notebooks.

**Note: the bpz branch install is a temporary measure allowing goldenspike to run with a version of bpz that uses the updated directory names**

Using bash loops to iterate through all creation/core/etc notebooks in the creation/core/etc step, to provide convenient grouping for more readable workflow outputs. This also means we only have to specify each notebook once, in the NOTEBOOKS environment variable.


drewoldag (approved):
Very cool. Thanks for putting this together!

PR #363, opened 2023-05-04
attempt to build a release for pypi

eacharles:


sschmidt23 (approved):
Looks good, and this should get cleaned up once dsps is in its own repo and/or the dsps updates get propagated into RAIL.

PR #368, opened 2023-05-08
Issue/355/cleanups

olivialynn:
For LSSTDESC/rail_hub#25, some very simple tidying.

After a talk with Alex, we decided it was better to go ahead and remove the goldenspike yaml, because the fixed paths they include can present difficulties, we already have a working example of reading in a yam pipeline file (Run_Pipe.ipynb), and running the goldenspike notebook will create a reusable pipeline file that works with the user's environment.


eacharles:
I'd like to proposed a different solution to the paths in the goldenspike notebook.yaml, which is to set the paths relative to $RAILDIR and to os os.path.expandvars in the rail/core/data.py functions that open files.   I think that is a better solution long term

tests/test_dsps_interface.py
delucchi-cmu, line 1: I kinda like that the tests directory mirrors the src directory in this instance, and I'd prefer if *more* of the tests reflected that structure (even if that means writing more tests so there are more files =]).

Is there some plan to flatten out the src directory? If not, I'm happy where these files currently live.

delucchi-cmu (commented):
Mmm. I love how much deletion is happening here.

olivialynn:
> I'd like to proposed a different solution to the paths in the goldenspike notebook.yaml, which is to set the paths relative to $RAILDIR and to os os.path.expandvars in the rail/core/data.py functions that open files. I think that is a better solution long term

Heard--I have a new PR open to address the pipeline file and get it functioning again, please let me know what you think.

I chose to set the yaml paths so they'll work if the user calls it from the root of their RAIL directory (I believe this is $RAILDIR - ie, the directory from where they'd have to call `ceci examples/goldenspike_examples/goldenspike.yml` to run it) for the time being, but interested in future path flexibility.

eacharles (approved):
Looks good, thanks.

PR #381, opened 2023-05-11
Issue/374/introspection

eacharles:


eacharles:
Main changes:

1. Added src/rail/core/introspection.py with a class called RailEnv that keeps track of packages and namespaces that are available in the rail ecosystem in the current python environment
2. Added src/rail/stages packages that uses RailEnv to dynamically populate itself with available RailStage objects
3. Added unit tests for same.

sschmidt23 (approved):
Looks good, seems to function as intended, thanks.

PR #382, opened 2023-05-16
Issue/347/mw extinction

hangqianjun:
Adding a function to the obs_condition module such that when an E(B-V) map is supplied, the MW extinction reddening is applied to the data for each band before passing on to the LSST error model.


eacharles (approved):
I'm going to approve this, but would you mind having a look at the dereddener class in src/rail/core/utilStages.py and thinking a bit about the overlaps?

I think it is fine to have both for now, but we should at least make sure the parameter names and usage is the same.  (Feel free to suggest changres to the Dereddener if you think that makes more sense.)

PR #389, opened 2023-06-05
Removed deprecated PIT and Brier metrics and unit tests

### LINKED ISSUE #388: Remove deprecated PIT metrics and unit tests
PIT metrics were moved to qp while working on issue #283 

It's time to remove the deprecated metrics and unit tests in RAIL.

drewoldag:
These metrics and tests were ported to qp several months ago. There don't seem to be any places where the RAIL implementations are still being used - they would have been producing deprecation warnings for months. Now they will simply fail, and we can update them as needed.


sschmidt23:
It looks like RAIL/examples/evaluation_examples/demo.ipynb
https://github.com/LSSTDESC/RAIL/blob/main/examples/evaluation_examples/demo.ipynb
still uses the deprecated package, can you update the references in there to point to qp rather than the old deprecated versions? I think that's the only demo that hasn't been updated.

drewoldag:
@sschmidt23 I think that we're all updated at this point. Let me know if you notice anything else!

aimalz (approved):
Looks good to me! @sschmidt23 Are your concerns now addressed?

sschmidt23:
Yep, everything looks good and runs for me now, thanks @drewoldag , I think this is fine to be merged.

Issue #384, opened 2023-05-22
Isolating things into new packages


olivialynn:
If there are things we want to consider isolating into new packages outside the minimal installation for rail, we should list them out here:

* somoclu - tricky dependencies that are not ideal for base rail, needs conda installation on Mac M1+'s
* dsps - stability questions?

eacharles:
Honestly I think that the direct dependecies for the base rail package ideally would be:  

deprecated
numpy
qp-prob  # (note, this is not qp-prob[full], and does not include h5py, panda, pyarrow, and astropy)
scipy
ceci

And indirectly, via qp-prob
tables-io  # (as above, this is not tables-io[full], and does not include h5py, panda, pyarrow, and astropy)

This results in about 75 packages.
Not that this doesn't actually include the code to read a specific file formats, but since different rail modules read different types of files, I think that is ok.

eacharles:
This would involve:

1. a package for astro tools kinda stuff with these addtional deps:
    "astropy",
    "healpy",
    "photerr",
    "dustmaps",
    "pz-hyperbolic-temp",

2. A package for dsps with these additional direct deps:
    "dsps",

3. A package for pzflow dependent stuff with these additional direct deps:
    "pzflow"

4. A package from som-dependent stuff, with these additional direct deps:
    "scikit-learn",
    "minisom",
    "somoclu",

5. a package from scikit-learn dependent stuff, with this addtional deps:
    "scikit-learn",

eacharles:
Given the above breakdown, here is where the files that are currently in RAIL might end up:

RAIL/src/rail/core/__init__.py                  base
RAIL/src/rail/core/_version.py                 base 
RAIL/src/rail/core/algo_utils.py               base
RAIL/src/rail/core/common_params.py    base
RAIL/src/rail/core/data.py                        base
RAIL/src/rail/core/introspection.py          base
RAIL/src/rail/core/stage.py                      base
RAIL/src/rail/core/utilPhotometry.py         ?
RAIL/src/rail/core/utilStages.py                 ?
RAIL/src/rail/core/utils.py                        base
RAIL/src/rail/creation/degradation/__init__.py  remove
RAIL/src/rail/creation/degradation/grid_selection.py    degredation
RAIL/src/rail/creation/degradation/lsst_error_model.py
RAIL/src/rail/creation/degradation/observing_condition_degrader.py
RAIL/src/rail/creation/degradation/quantityCut.py
[...]

PR #391, opened 2023-06-09
Issue/288/standardize naming

aimalz:
This is a draft PR for LSSTDESC/rail_base#1  so I can check that I'm not breaking the tests (because fitsio isn't behaving on my local machine today). The scope of changes is to standardize naming of classes (and hopefully their config parameters, although I'm not sure I understand them all well enough to do that alone) that are going to be isolated into standalone repos in LSSTDESC/RAIL#384, to be accompanied by PRs in the existing standalone repos as necessary. To prevent myself from getting distracted from the task at hand, I'm also adding `TODO` notes to give concrete starting points for LSSTDESC/rail_hub#22.

EDIT: _Overall, the changes here pertain solely to consistency in class and module naming_; apologies if this was unclear to anyone! The purpose is so we don't inadvertently ask users to guess at the syntax for a given stage with respect to the module it's in (occasionally risking namespace issues if they import common packages with the same as a RAIL module) and with respect to other stages that do similar things with other underlying algorithms, across `creation` and `evaluation` as well as within `estimation` (leaving open the possibility to distinguish algorithms that can be used for both estimation and summarization). It also addresses a common source of confusion among users trying to understand the idea of stages in general, due to naming standards we established before switching to ceci as a back-end: rather than being a function or method, if a stage is an object that takes an extra line of code to run, why are they sometimes named as actions or underlying algorithms (without indication of whether it's an estimator or summarizer)?

In any case, **this PR affects formatting only, not content/meaning of any of the classes in question**, so my hope is that the disruption is entirely degenerate with that anticipated for LSSTDESC/RAIL#384, if they happen at the same time. The changes include the following (just in rail.estimation for now):

- [x] delightPZ.py --> delight_hybrid.py
- [x] Inform_DelightPZ --> DelightInformer
- [x] delightPZ --> DelightEstimator
- [x] GPz.py --> _gpz_util.py
- [x] Inform_GPz_v1 --> GPzInformer
- [x] GPz_v1 --> GPzEstimator
- [x] knnpz.py --> k_nearneigh.py
- [x] Inform_KNearNeighPDF --> KNNInformer --> KNearNeighInformer
- [x] KNearNeighPDF --> KNNEstimator --> KNearNeighEstimator
- [x] naiveStack.py --> naive_stack.py
- [x] NaiveStack --> NaiveStackSummarizer
- [x] Inform_NZDir --> NZDirInformer
- [x] NZDir.py --> nz_dir.py
- [x] NZDir --> NZDirSummarizer
- [x] pointEstimateHist.py --> point_est_hist.py ?--> point_est_dist.py
- [x] PointEstimateHist --> PointEstHistSummarizer ?--> PointEstDistSummarizer
- [x] Inform_PZFlowPDF --> PZFlowInformer
- [x] pzflow.py --> pzflow_nf.py
- [x] PZFlowPDF --> PZFlowEstimator
- [x] randomPZ.py --> random_gauss.py ?--> rand_gauss.py
- [x] RandomPZ --> RandomGaussEstimator ?--> RandGaussEstimator
- [x] simpleSOM.py --> minisom_som.py
- [x] Inform_SimpleSOMSummarizer --> MiniSOMInformer
- [x] SimpleSOMSummarizer --> MiniSOMSummarizer
- [x] sklearn_nn.py --> simpleNN.py --> skl_neurnet.py
- [x] Inform_SimpleNN --> SimpleNNInformer --> SklNeurNetInformer
- [x] SimpleNN --> SimpleNNEstimator --> SklNeurNetEstimator
- [x] somocluSOM.py --> somoclu_som.py
- [x] Inform_somocluSOMSummarizer --> SOMocluInformer
- [x] somocluSOMSummarizer --> SOMocluSummarizer
- [x] trainZ.py --> train_z.py
- [x] Inform_trainZ --> TrainZInformer
- [x] TrainZ --> TrainZEstimator
- [x] varInference.py --> var_inf.py
- [x] VarInferenceStack --> VarInfStackSummarizer

I can put together a block of `import X as Y` for backwards compatibility in case anyone finds the changes too hard to propagate through pipelines you've already written.

Note that the base classes are unaffected (though they will be in LSSTDESC/rail_base#14, at least there aren't any Evaluators outside base RAIL), so stages that are not updated for consistency won't be broken. Propagation of consistent naming to the already standalone repos will be completed in subsequent PRs in those repos. However, I'd like to take advantage of the opportunity posed by the isolation-refactoring disruption to make similar changes to some of the degraders, on a separate issue/branch/PR akin to LSSTDESC/rail_base#14.


aimalz:
@yanzastro @sschmidt23 While propagating naming consistency through the demos I noticed that there are [two seemingly slightly different versions of the somoclu demo](https://github.com/LSSTDESC/RAIL/tree/main/examples/estimation_examples). Can you confirm which I should remove (or update the markdown text to indicate the big picture differences I failed to find)?

EDIT: Related question for @sschmidt23 : Since both SimpleSOM and somocluSOM contained the same `_computemagcolordata` function, I moved it to `src/rail/core/algo_utils.py`. Then I discovered more very similar functions, `make_color_data` in sklearnNN and `_computecolordata` in knnpz. I think these are equivalent so all the algos using this functionality could call a single version of this function in  `src/rail/core/algo_utils.py` (or possibly in a new `src/rail/estimation/utils.py` module instead) but they are indeed calculating the color-plus-one-mag information differently. Can you advise on whether this assessment is accurate and, if so, which version would be preferable?

